PR #36, opened 2018-03-30
Use singleton-like Meta class to store global state

### LINKED ISSUE #34: Eliminate global variable use from meta.py
Right now, snorkel uses some hacky global variables in [meta.py](https://github.com/HazyResearch/fonduer/blob/master/fonduer/snorkel/models/meta.py). This poses two main issues:

1. It forces the user to set the `SNORKELDB` environment variable _before_ importing
2. It is executing code on import alone. For example, just importing `fonduer` will create a `snorkel.db` file immediately. We should [say no to import side-effects](https://chrismorgan.info/blog/say-no-to-import-side-effects-in-python.html).

Perhaps this should be some sort of Session class instead, with attributes that are accessed when needed, rather than using these global variables throughout the codebase.

lukehsiao:
Closes #34.


lukehsiao:
This release is NOT backwards compatable with v0.1.4. Specifically, in order
to initialize a session with postgresql, you no longer do
```py       
os.environ['SNORKELDB'] = 'postgres://localhost:5432/' + DBNAME
from fonduer import SnorkelSession       
session = SnorkelSession()
```

which had the side-effects of manipulating your database tables on import (or creating a `snorkel.db` file if you forgot to set the environment variable). Now, you use the Meta class to initialize your session:

```py
from fonduer import Meta       
session = Meta.init("postgres://localhost:5432/" + DBNAME).SnorkelSession()
```
      
No side-effects occur until ``Meta`` is initialized.

tests/e2e/test_e2e.py
```diff
@@ -37,15 +35,14 @@
...
-    'CI' not in os.environ, reason="Only run e2e on Travis")
+@pytest.mark.skipif('CI' not in os.environ, reason="Only run e2e on Travis")
 def test_e2e(caplog):
     """Run an end-to-end test on 20 documents of the hardware domain."""
     caplog.set_level(logging.INFO)
     PARALLEL = 2
     max_docs = 12
 
-    session = SnorkelSession()
+    session = Meta.init('postgres://localhost:5432/' + DB).SnorkelSession()
```
SenWu: Can we update the name here (FonduerSession)?

tests/parser/test_parser.py
```diff
@@ -29,6 +25,7 @@ def test_parse_structure(caplog):
     """
     caplog.set_level(logging.INFO)
     logger = logging.getLogger(__name__)
+    session = Meta.init('postgres://localhost:5432/' + ATTRIBUTE).SnorkelSession()
```
SenWu: Same here.

tests/parser/test_parser.py
```diff
@@ -77,7 +74,7 @@ def test_parse_document_md(caplog):
     This tests both the structural and visual parse of the document.
     """
     logger = logging.getLogger(__name__)
-    session = SnorkelSession()
+    session = Meta.init('postgres://localhost:5432/' + ATTRIBUTE).SnorkelSession()
```
SenWu: Same here.

tests/parser/test_parser.py
```diff
@@ -126,7 +123,7 @@ def test_parse_document_diseases(caplog):
     """
     caplog.set_level(logging.INFO)
     logger = logging.getLogger(__name__)
-    session = SnorkelSession()
+    session = Meta.init('postgres://localhost:5432/' + ATTRIBUTE).SnorkelSession()
```
SenWu: Same here.

docs/index.rst
```diff
@@ -18,20 +18,21 @@ submit your contributions as a pull request.
...
 .. toctree::
-   :maxdepth: 2
-   :caption: Developer Documentation
-
-   dev/changelog
-   dev/install
-   dev/tests
+    :maxdepth: 2
+    :caption: Developer Documentation
+
```
SenWu: remove the extra spaces

PR #44, opened 2018-04-03
Switch to logging, absorb Snorkel, and remove SQLite

### LINKED ISSUE #27: Switch to logging in Fonduer.

### LINKED ISSUE #29: Move all snorkel code directly into Fonduer.
Remove snorkel subdirectory.

The motivation here is that the line between importing from snorkel or fonduer is a little blurry, and kind of unnecessary. It may simplify the code if we just absorb all the snorkel files into fonduer directly.

### LINKED ISSUE #35: Scrub code related to sqlite, which is unsupported in fonduer
In Fonduer we rely on parallelization to improve performance by using postgres. We do not support sqlite. Removing any code that relates to sqlite will simplify the code base and reduce confusion.

lukehsiao:
Closes #29, closes #27, closes #35.


fonduer/parser.py
```diff
@@ -21,6 +22,34 @@
 logger = logging.getLogger(__name__)
 
 
+class PDFPreprocessor(DocPreprocessor):
```
lukehsiao: I'm also taking in this preprocessor from new_parser branch, even though it is currently unused in the tutorials.

fonduer/parser/parser.py
```diff
@@ -0,0 +1,53 @@
...
+        '''
+        if sys.version_info[0] < 3:
+            text_alt = text.encode('utf-8', 'error')
+            text_alt = text_alt.decode('string_escape', errors='ignore')
+            text_alt = text_alt.decode('utf-8')
+            return text_alt
+        else:
+            return text
+
+    def connect(self):
```
SenWu: Do we need this function?

fonduer/parser/parser.py
```diff
@@ -0,0 +1,53 @@
...
+        '''
+        Return connection object for this parser type
+        :return:
+        '''
+        raise NotImplemented
+
+    def close(self):
```
SenWu: Same

fonduer/parser/parser.py
```diff
@@ -0,0 +1,53 @@
...
+        '''
+        Kill this parser
+        :return:
+        '''
+        raise NotImplemented
+
+
+class ParserConnection(object):
```
SenWu: Same here?

lukehsiao:
Updated tutorials pushed.

PR #45, opened 2018-04-04
Add matchers, preprocessors, data_model to docs

### LINKED ISSUE #43: Add data model to docs
We need better docs for the data model. For example, given a candidate in the lf_helpers, how do I view the sentence one of it's mentions is it? What attributes does a `Span` have? These are frequent questions that can be answered by looking at the code directly, but would be much more user friendly if we had docs for them.

Related: #32

lukehsiao:
Closes #43.

Of course, these docs could be more complete, but this sets up the files for later expansion.

### DIFF
--- a/tests/e2e/test_e2e.py
+++ b/tests/e2e/test_e2e.py
@@ -23,7 +23,6 @@ ATTRIBUTE = "stg_temp_max"
 DB = "e2e_test"
 
 
-
 @pytest.mark.skipif('CI' not in os.environ, reason="Only run e2e on Travis")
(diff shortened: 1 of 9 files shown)


tests/e2e/test_e2e.py
```diff
@@ -52,7 +51,7 @@ def test_e2e(caplog):
     logger.info("Phrases: {}".format(num_phrases))
     #  assert num_phrases == 20
 
-    # Divid into test and train
+    # Divide into test and train
```
SenWu: good catch!

PR #52, opened 2018-05-14
Extending styles parsing and RegEx search

Prabh06:
Extended two features:

- Added regex search and toggle for full span match in RegexMatch class.

- Added styles parsing from the style class and appended to existing html styles attr.


lukehsiao:
Working on getting travis tests fixed: https://github.com/HazyResearch/numbskull/pull/54.

fonduer/matchers.py
```diff
@@ -228,11 +228,18 @@ def init(self):
         self.attrib = self.opts.get('attrib', WORDS)
         self.sep = self.opts.get('sep', " ")
 
+        # Extending the RegexMatch to handle search(instead of only match) 
+        # and adding a toggle for full span match.
+        # Default values are set to False and True for search flag and full
```
lukehsiao: Please get rid of these trailing whitespaces.

Make sure the code still passes our code style check: `make check`

Prabh06: Done

fonduer/parser/parser.py
```diff
@@ -260,6 +260,28 @@ def parse_node(node, table_info=None, figure_info=None):
...
+                                                exp = r'(.' + x[1] + ')([\n\s\r]*)\{(.*?)\}'
+                                                r = re.compile(exp, re.DOTALL)
+                                                if r.search(styles.text) is not None:
+                                                    if cur_style_index is not None:
+                                                        parts['html_attrs'][cur_style_index] += r.search(styles.text).group(3)\
+                                                            .replace('\r', '').replace('\n', '').replace('\t', '')
+                                                    else:
+                                                        parts['html_attrs'] = 'style=' + r.search(styles.text).group(3)\
+                                                            .replace('\r', '').replace('\n', '').replace('\t', '')
+                                                break
```
lukehsiao: It's not super obvious to me what the use case for this code is. It looks to me like if an element has style attributes, then your extending the element's styles with the first style class in the head of the document that you find, and only the first style class. Is that correct? Could you explain why you needed this code?

Also, could you add a test case for this code?  This would be added to [test_parser.py](https://github.com/HazyResearch/fonduer/blob/master/tests/parser/test_parser.py).

The easiest way to do this would be to add a new, simplified HTML document into `tests/data/html/` (you can create this new html directory). Then just add a case similar to:

```py
def test_parse_style(caplog):
    """Test style tag parsing."""
    caplog.set_level(logging.INFO)
    session = Meta.init('postgres://localhost:5432/' + ATTRIBUTE).Session()

    max_docs = 1
    docs_path = 'tests/data/html/[your new document].html'

    # Preprocessor for the Docs
    preprocessor = HTMLPreprocessor(docs_path, max_docs=max_docs)

    # Grab the document, text tuple from the preprocessor
    doc, text = next(preprocessor.generate())
    logger.info("    Text: {}".format(text))

    # Create an OmniParserUDF
    omni_udf = OmniParserUDF(
        True,           # structural
        [],             # blacklist, empty so that style is not blacklisted
        ["span", "br"], # flatten
        '',             # flatten delim
        True,           # lingual
        True,           # strip
        [],             # replace
        True,           # tabular
        True,           # visual
        pdf_path,       # pdf path
        Spacy())        # lingual parser

    # Grab the phrases parsed by the OmniParser
    phrases = list(omni_udf.parse_structure(doc, text))

    # Add your assertions
```

Prabh06: If an html element `<elm class='.s2'>` has a class attribute `here .s2` then this code will find that class `s2` in `<style>`  and append/assign the `styles` defined in the `s2` class to  style attribute in `parts['html_attrs']`

fonduer/matchers.py
```diff
@@ -242,12 +249,18 @@ def _f(self, c):
...
+    """
+        Matches regex pattern on **full concatenated span**
+        If search flag is set to True:
+            Search regex pattern in **full concatenated span**
+    """
     def _f(self, c):
-        return True if self.r.match(
-            c.get_attrib_span(self.attrib,
-                              sep=self.sep)) is not None else False
+        if not self.search:
```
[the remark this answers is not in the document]
Prabh06: Yes, done.

fonduer/parser/parser.py
Prabh06: Will add test case and check in

fonduer/matchers.py
```diff
@@ -228,11 +228,18 @@ def init(self):
         self.attrib = self.opts.get('attrib', WORDS)
         self.sep = self.opts.get('sep', " ")
 
+        # Extending the RegexMatch to handle search(instead of only match)
+        # and adding a toggle for full span match.
+        # Default values are set to False and True for search flag and full
+        # span matching flag respectively.
+        self.search = self.opts.get('search', False)
+        self.full_match = self.opts.get('full_match', True)
```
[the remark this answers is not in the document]
Prabh06: `self.full_match` is to toggle appending `$` to regex

Eg:

```
phrase = 'Invoice#:2387621387'
r1 = re.compile(r'Invoice')
r2 = re.compile(r'(Invoice)$')

r1.search(phrase) # returns <_sre.SRE_Match object; span=(0, 7), match='Invoice'>
r2.search(phrase) # returns None
```
This is happening because $ matches the end of the string but the expression can be part of the span not at the end.

Prabh06:
@lukehsiao  added test case. Please review

PR #55, opened 2018-05-23
remove viewer

SenWu:


fonduer/viewer.py
```diff
@@ -1,349 +0,0 @@
-from __future__ import (absolute_import, division, unicode_literals)
-
-import getpass
-import os
-import warnings
-
-import ipywidgets as widgets
```
lukehsiao: What dependencies can we remove if we remove the Viewer? Seems like we don't need ipywidget dependency anymore?

PR #60, opened 2018-06-30
Split sqlalchemy models into individual files

### LINKED ISSUE #57: Break up large Python files into reasonable sub-files
For example, context.py is over 1000 lines of code and contains many classes which could reasonably be made into separate files.

lukehsiao:
Closes #57.


fonduer/parser/models/__init__.py
```diff
@@ -1,20 +1,16 @@
-from fonduer.parser.models.context import (
-    Cell,
-    Context,
-    Document,
-    Figure,
-    Image,
+from fonduer.parser.models.context import Context, construct_stable_id, split_stable_id
+from fonduer.parser.models.document import Document
+from fonduer.parser.models.figure import Figure
+from fonduer.parser.models.image import Image, TemporaryImage
```
SenWu: This is for candidate generation?

fonduer/parser/models/__init__.py
```diff
@@ -1,20 +1,16 @@
...
+from fonduer.parser.models.phrase import Phrase
+from fonduer.parser.models.span import (
```
SenWu: This is for candidate generation?

lukehsiao:
Following up on our conversation, will move the classes that are specific to candidate generation, and not core to the data model, into the candidate generation phase (e.g., Image, ImplicitSpan).

PR #62, opened 2018-07-05
update osx test

SenWu:


lukehsiao:
Right now we have the `deploy` step upload the latest release to PyPi when we make a tag and it passes on Python 3.6. We only want it to deploy to PyPi once, or the subsequent deploys will fail. Do we need to modify the `deploy` so that it only deploys on Linux with Python 3.6?

Issue #56, opened 2018-06-22
Refactor codebase into submodules for each pipeline phase


lukehsiao:
Currently, Fonduer is kind of a monolitic package, where all database tables are created on init. In order to make development easier, we want to split Fonduer into sobmodules, each of which a single task in the pipeline:

1. parser
2. candidates
3. featurization
4. supervision
5. learning
6. utils

Database tables will only be created in the initialization of each of these independent modules.

TODO:
- [x] Reorganize files as is, just fixing imports and setting up new directories
- [x] Use all absolute imports
- [ ] ~~Split up the initialization performed by Meta into the respective pipeline phases~~
- [x] Define each phase's API. Expose submodules, rather than all functions, in the root fonduer package
- [x] Simplify the source files where possible

lukehsiao:
All of our tables inherit from `delcarative_base()` so we should be able to do something similar to [[1]](https://stackoverflow.com/questions/19175311/how-to-create-only-one-table-with-sqlalchemy):

```
[class].__table__.create(_meta.engine)
```

We probably want to use `checkfirst=True` [[2]](http://docs.sqlalchemy.org/en/latest/core/metadata.html?highlight=create_all#creating-and-dropping-database-tables).

lukehsiao:
We need to do some more cleaning-up. For example, we currently name our basic unit "Phrases" since historically, we were based on Snorkel which already took the name "Sentences". Now that we have rewritten this code, we should switch back to "Sentences".

lukehsiao:
We also want classes for `section`, `row`, `col`, `caption`, and `paragraph`.

PR #64, opened 2018-07-07
Split doc_preprocessors into separate files

### LINKED ISSUE #63: Break doc_preprocessors into a directory of individual files
Rather than having all the preprocessors in a single file.

lukehsiao:
Closes #63.
Related to #56.


Makefile
```diff
@@ -7,9 +7,15 @@ dev:
...
-check: dev 
+check:
 	isort -rc -c fonduer/
 	black fonduer/ --check
+	# This is our code-style check. We currently allow the following exceptions:
+	#   - E731: do not assign a lambda expression, use a def
+	#   - W503: line break before binary operator
+	#   - E741: do not use variables named ‘l’, ‘O’, or ‘I’
+	#   - E203: whitespace before ‘:’
+	flake8 fonduer --count --max-line-length=127 --statistics --ignore=E731,W503,E741,E203
```
lukehsiao: Added flake8 check back into the makefile because it actually makes refactoring a little easier (e.g. highlights modules that are imported and unused or not imported)

fonduer/parser/doc_preprocessors.py
```diff
@@ -1,237 +0,0 @@
...
-                    # if column is set, retrieve specific column from CSV record
-                    doc_path = doc_path.split(self.delim)[self.column]
-                yield doc_path.strip()
-
-    def parse_file(self, fp, file_name):
-        return self.parser.parse_file(fp, file_name)
-
-
-#  NOTE: this is not used in Fonduer
-#  class HTMLDocPreprocessor(DocPreprocessor):
```
lukehsiao: Note that previously, when we were just directly using snorkel, we needed to avoid overloading this name so we made `HTMLPreprocessor`. Now that we don't use this, I've switched to use Fonduer's version and called it `HTMLDocPreprocessor` for naming consistency with the other preprocessors.

PR #76, opened 2018-07-18
Add Paragraph and Section Contexts

lukehsiao:


fonduer/parser/parser.py
```diff
@@ -269,13 +283,117 @@ def _parse_figure_node(self, node, state):
 
         return state
 
-    def _parse_sentence(self, node, state):
+    def _parse_sentence(self, paragraph, node, text, field, state):
```
lukehsiao: I'm not a big fan of how many arguments this function has. Any suggestions on simplifying this would be welcome. I don't want to have to alter the `state`.

tests/data/html_simple/md_para.html
```diff
@@ -0,0 +1,62 @@
+<html lang="en" xml:lang="en" xmlns="http://www.w3.org/1999/xhtml">
+
+<head>
+</head>
+
+<body>
+    <h1 id="sample-markdown">Sample Markdown</h1>
+    <p>This is some basic, sample markdown. Unlike the other markdown document, however, this document actually contains paragraphs of text. That is, larger amounts of text that are all present in a single HTML node like the one you are currently reading.</p>
```
lukehsiao: Just added this document so that we have a real paragraph to test.

fonduer/parser/models/table.py
```diff
@@ -17,80 +17,20 @@ class Table(Context):
...
 
     def __repr__(self):
-        return "Table(Doc: {}, Position: {})".format(self.document.name, self.position)
-
-    def __gt__(self, other):
-        # Allow sorting by comparing the string representations of each
-        return self.__repr__() > other.__repr__()
-
-
-class Row(Context):
```
lukehsiao: Dropping Row/Col

Makefile
```diff
@@ -10,13 +10,15 @@ test: dev check
 
 check:
 	isort -rc -c fonduer/
```
lukehsiao: Even if we don't enforce flake8, add style checks for the tests.

SenWu: 👍

fonduer/parser/models/sentence.py
```diff
@@ -200,8 +212,12 @@ def __repr__(self):
                 self.text,
             )
         else:
-            return "Sentence (Doc: '{}', Index: {}, Text: '{}')".format(
-                self.document.name, self.sentence_num, self.text
+            return "Sentence (Doc: '{}', Sec: {}, Par: {}, Index: {}, Text: '{}')".format(
+                self.document.name,
+                self.section.position,
+                self.paragraph.position,
+                self.sentence_num,
```
SenWu: Can you all use position instead of some position and some num?

fonduer/parser/parser.py
```diff
@@ -143,25 +144,33 @@ def apply(self, x, **kwargs):
...
+            # Create the Table in the DB
+            parent = state["parent"][node]
+            if not isinstance(parent, Section):
+                logger.warning("Table is nested within {}".format(parent))
             state["context"][node] = Table(
-                document=state["document"], stable_id=stable_id, position=table_idx
+                document=state["document"],
+                # TODO: This just takes the one and only Section in a document
+                # and assigns it as the Table's parent.
+                section=state["document"].sections[0],
```
SenWu: why sections[0]?

fonduer/parser/parser.py
```diff
@@ -288,100 +406,62 @@ def _parse_sentence(self, node, state):
...
-                    abs_sentence_offset_end,
+            # Process the Paragraph
+            stable_id = "{}::{}:{}".format(
+                state["document"].name, "paragraph", state["paragraph"]["idx"]
+            )
+            parts = {}
+            parts["stable_id"] = stable_id
+            parts["document"] = state["document"]
+            parts["position"] = state["paragraph"]["idx"]
+            parts["section"] = state["document"].sections[0]
```
SenWu: same here. why sections[0] here instead of the current section from parents?

fonduer/parser/parser.py
```diff
@@ -288,100 +406,62 @@ def _parse_sentence(self, node, state):
...
-                    + parts["char_offsets"][-1]
-                    + len(parts["words"][-1])
-                )
-                parts["stable_id"] = construct_stable_id(
-                    document,
-                    "sentence",
-                    state["sentence"]["abs_offset"],
-                    abs_sentence_offset_end,
+            # Process the Paragraph
+            stable_id = "{}::{}:{}".format(
```
SenWu: Can we use the same format for stable_id?

fonduer/parser/parser.py
```diff
@@ -269,13 +283,117 @@ def _parse_figure_node(self, node, state):
...
+            )
+            parts["stable_id"] = construct_stable_id(
+                document,
+                "sentence",
+                state["sentence"]["abs_offset"],
+                abs_sentence_offset_end,
+            )
+            state["sentence"]["abs_offset"] = abs_sentence_offset_end
+            if self.structural:
+                context_node = node.getparent() if field == "tail" else node
```
SenWu: I know this is old code. Just make sure the logic here is correct. Will the the tail node be the sibling of the text node or now?

fonduer/parser/parser.py
SenWu: Can we put text and field into state? Then when we done with _parse_sentence we can remove those two.

fonduer/parser/models/table.py
SenWu: 👍

tests/parser/test_parser.py
```diff
@@ -103,6 +105,74 @@ def test_parse_md_details(caplog):
     assert header.dep_labels == ["compound", "ROOT"]
 
 
+def test_parse_md_paragraphs(caplog):
```
SenWu: Add some checks for logic like some sentence's parent is some paragraph, section etc..

fonduer/parser/parser.py
lukehsiao: Yes. Good idea.

fonduer/parser/parser.py
lukehsiao: What do you mean?

fonduer/parser/parser.py
lukehsiao: So to expand on this a little. Paragraph can be in Section, Table, or Cell. Would need to add several lines of logic to navigate up to the Section.

tests/parser/test_parser.py
lukehsiao: Sentence can only have a Paragraph parent.

SenWu:
LGTM.

PR #78, opened 2018-07-20
Add basic figure and figcaption support

lukehsiao:
This adds basic caption support for `<figure>` elements, as well as
support for `<figure>` elements themselves. Two important limitations of
the current implementation is:

- No support for nested <figure> elements
- No support for <figure> elements with multiple  elements


tests/parser/test_parser.py
```diff
@@ -76,7 +76,7 @@ def test_parse_md_details(caplog):
 
     # Check that doc has sentences
     assert len(doc.sentences) == 45
-    sent = doc.sentences[25]
+    sent = sorted(doc.sentences, key=lambda x: x.position)[25]
```
lukehsiao: Noticed some inconsistency in sorting this time around, so just sorting the sentences based on their position before testing so that it is deterministic.

fonduer/parser/parser.py
```diff
@@ -262,28 +262,53 @@ def _parse_figure(self, node, state):
...
+                position=state["figure"]["idx"],
+                url=node.get("src"),
+            )
+        elif node.tag == "figure":
+            # Pull the image from a child img node, if one exists
+            imgs = [child for child in node if child.tag == "img"]
+
+            if len(imgs) > 1:
+                logger.warning("Figure contains multiple images.")
+                raise NotImplementedError("No support for figures with multiple <img>")
```
lukehsiao: Right now, this just crashes. If we want to be more robust, a better answer might be to basically strip the `<figure>` tag and process all the included images individually. Not sure what we'd do about the a `<figcaption>` if one was present. Perhaps just ignore it.

SenWu:
LGTM.

PR #81, opened 2018-07-21
reorg_lf_helpers

SenWu:


fonduer/supervision/lf_helpers.py
```diff
@@ -1,125 +1,9 @@
...
-        getattr(span.sentence, attrib)[i + 1 : i + 1 + window],
-        n_min=n_min,
-        n_max=n_max,
-        lower=lower,
-    ):
-        yield ngram
+from fonduer.supervision.utils.structural import *
+from fonduer.supervision.utils.tabular import *
+from fonduer.supervision.utils.textual import *
+from fonduer.supervision.utils.visual import *
```
lukehsiao: I think the correct way to organize this so that the imports to the user is the same is to make lf_helpers the directory with an __init__.py, and having the separate modaility helper files in that directory, rather than making a utils directory. If that sounds OK to you, I can take care of that update.

SenWu: That's sound right! Thanks!

fonduer/supervision/lf_helpers/__init__.py
```diff
@@ -0,0 +1,147 @@
...
+    get_horz_ngrams,
+    get_vert_ngrams,
+    get_page_vert_percentile,
+    get_page_horz_percentile,
+    get_visual_aligned_lemmas,
+    get_aligned_lemmas,
+)
+
+
+def is_superset(a, b):
```
lukehsiao: This function and overlap() seemed too general to be in tabular.py, so I moved them here.

fonduer/supervision/lf_helpers/visual.py
```diff
@@ -0,0 +1,433 @@
...
+    return
+
+
+def get_vert_ngrams_center(c):
+    """Not implemented."""
+    # TODO
+    return
+
+
+def get_visual_header_ngrams(c, axis=None):
```
lukehsiao: I don't expose these unimplemented helpers in the __init__.py

fonduer/supervision/lf_helpers/__init__.py
SenWu: 👍

fonduer/supervision/lf_helpers/visual.py
SenWu: 👍

PR #93, opened 2018-08-08
Add Mentions, and generate Candidates based on Mentions

lukehsiao:
This also includes some improvements to tests and a bugfix for the Meta class.

PR #95, opened 2018-08-09
Fix visualizer to support new Candidate/Mention structure

lukehsiao:
Updates the visualizer so that it still works after #93.


SenWu:
LGTM.

Issue #90, opened 2018-08-02
Transistor Image Tutorial - UnicodeEncodeError from CorpusParser


Allen8838:
Hello,

First off, thanks for this interesting library. I have only gone through it a little bit but am quite excited to explore its full capabilities.  

I am going through the transistor_image_tutorial at the moment. Running the lines:

```
corpus_parser = Parser(structural=True, lingual=True, visual=True, pdf_path=pdf_path, flatten=[])
%time corpus_parser.apply(doc_preprocessor, parallelism=PARALLEL)
```
I get the following UnicodeEncodeError. 

![transistor_image_tutorial - unicodeencodeerror](https://user-images.githubusercontent.com/35154803/43555578-05f25406-95c9-11e8-886d-d9b4a7dc049c.PNG)

As a result, when I execute the next line, I get 

Documents: 0
Sentences: 0
Figures: 0

I was wondering if there is a workaround for this or if I needed to do something first to avoid the error and get the same result as the original transistor_image_tutorial file. Thank you.

lukehsiao:
Hi @Allen8838, are you using OSX or Linux?  Also, is this the [latest tutorials](https://github.com/HazyResearch/fonduer-tutorials) using Fonduer 0.2.3?

Allen8838:
Hi @lukehsiao, I am using Bash on Ubuntu on Windows. Yes, this is the latest tutorial using Fonduer 0.2.3.

lukehsiao:
I'm not able to reproduce this on my Ubuntu machine. Can you post the full error log?

Allen8838:
Please see below. It's one error log, broken into several snips. 

![image](https://user-images.githubusercontent.com/35154803/43559087-797bdf62-95da-11e8-8bd1-da5e0815eee6.png)

![image](https://user-images.githubusercontent.com/35154803/43559246-50d023c4-95db-11e8-95cd-acd99fa4f925.png)

![image](https://user-images.githubusercontent.com/35154803/43559271-6717547c-95db-11e8-829f-8a8f55240ffe.png)

![image](https://user-images.githubusercontent.com/35154803/43559283-7974203c-95db-11e8-9428-ddaf26083480.png)

![image](https://user-images.githubusercontent.com/35154803/43559301-91bfb8a4-95db-11e8-90e2-81d072617f18.png)

![image](https://user-images.githubusercontent.com/35154803/43559314-a053a8f8-95db-11e8-9149-68d961e5590d.png)

[...]

PR #98, opened 2018-08-14
Update dev docs, remove udf reducer, add default encoding to meta.py

lukehsiao:
Adding the client encoding may help prevent issues like #90 [[ref]](http://docs.sqlalchemy.org/en/latest/dialects/postgresql.html#unicode-with-psycopg2).

### DIFF
--- a/fonduer/meta.py
+++ b/fonduer/meta.py
@@ -17,7 +17,12 @@ def new_sessionmaker():
     # Otherwise any e.g. query starts a transaction, locking tables... very
     # bad for e.g. multiple notebooks open, multiple processes, etc.
     if Meta.postgres and Meta.ready:
-        engine = create_engine(Meta.conn_string, isolation_level="AUTOCOMMIT")
+        engine = create_engine(
+            Meta.conn_string,
+            client_encoding="utf8",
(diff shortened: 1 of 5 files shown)


fonduer/meta.py
```diff
@@ -17,7 +17,12 @@ def new_sessionmaker():
     # Otherwise any e.g. query starts a transaction, locking tables... very
     # bad for e.g. multiple notebooks open, multiple processes, etc.
     if Meta.postgres and Meta.ready:
-        engine = create_engine(Meta.conn_string, isolation_level="AUTOCOMMIT")
+        engine = create_engine(
+            Meta.conn_string,
+            client_encoding="utf8",
+            use_batch_mode=True,
```
SenWu: do you still need this?

lukehsiao: It's not currently utilized since we're not using the [bulk execution functions](http://docs.sqlalchemy.org/en/latest/faq/performance.html#i-m-inserting-400-000-rows-with-the-orm-and-it-s-really-slow) from sqlalchemy. But, having it hear leaves that option open in the future.

PR #100, opened 2018-08-16
Add multiprocess progress bar

lukehsiao:
Also reorganizes the pytest opts.

Inspired by Catalin Voss's PR for Snorkel: https://github.com/HazyResearch/snorkel/pull/1000


tests/e2e/test_e2e.py
```diff
@@ -77,7 +77,7 @@ def test_e2e(caplog):
         corpus_parser = Parser(
             structural=True, lingual=True, visual=True, pdf_path=pdf_path
         )
-        corpus_parser.apply(doc_preprocessor, parallelism=PARALLEL)
+        corpus_parser.apply(doc_preprocessor, parallelism=PARALLEL, count=max_docs)
```
lukehsiao: Note that because `doc_preprocessor` is a generator, it has no length attribute to use to calculate the progress bar. Thus, the user either needs to pass in `max_docs` as the count (recommended, since it doesn't materialize everything) or wrap the `doc_preprocessor`: `corpus_parser.apply(list(doc_preprocessor), parallelism=PARALLEL)`.

tests/pytest.ini
```diff
@@ -1,3 +1,4 @@
 [pytest]
+addopts = -v -rsXx
```
lukehsiao: Sets the default commandline options for pytest so that we don't have to specify it multiple places.

tests/shared/hardware_utils.py
```diff
@@ -47,10 +48,8 @@ def load_hardware_labels(
     gold_dict = get_gold_dict(filename, attribute=attrib)
     cand_total = len(candidates)
     logger.info("Loading {} candidate labels".format(cand_total))
-    pb = ProgressBar(cand_total)
     labels = []
-    for i, c in enumerate(candidates):
-        pb.bar(i)
+    for i, c in enumerate(tqdm(candidates)):
```
lukehsiao: Note that this order is actually important [[1]](https://github.com/tqdm/tqdm#faq-and-known-issues). The `enumerate` should come around the `tqdm` call.

setup.cfg
```diff
@@ -1,9 +1,6 @@
 [aliases]
 test=pytest
 
-[tool:pytest]
-addopts = -v -rsXx
-
```
lukehsiao: These were being ignored due to the existence of pytest.ini in `tests/`.

tests/pytest.ini
SenWu: 👍

setup.cfg
SenWu: 👍

tests/shared/hardware_utils.py
SenWu: 👍

tests/e2e/test_e2e.py
SenWu: there is no way to get this value from doc_preprocessor?

PR #103, opened 2018-08-17
Fix pdf path validation in Parser

lukehsiao:
Also adds debug-level logging for calls to poppler-utils.

### DIFF
--- a/fonduer/parser/parser.py
+++ b/fonduer/parser/parser.py
@@ -2,6 +2,7 @@ import itertools
 import logging
 import os
 import re
+import warnings
 from builtins import range
 from collections import defaultdict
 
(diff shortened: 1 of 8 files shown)


fonduer/utils/udf.py
```diff
@@ -9,7 +9,7 @@
 
     if "IPKernelApp" not in get_ipython().config:
         raise ImportError("console")
-except ImportError:
+except (AttributeError, ImportError):
```
lukehsiao: Needed to catch this error as well to run tests locally.

fonduer/parser/visual_linker.py
```diff
@@ -45,13 +45,21 @@ def parse_visual(self, document_name, sentences, pdf_path):
             yield sentence
 
     def extract_pdf_words(self):
+        self.logger.debug(
+            "pdfinfo '{}' | grep -a Pages | sed 's/[^0-9]*//'".format(self.pdf_file)
+        )
```
lukehsiao: This way, we can switch to debug level logging and get an exact log of the command that was run.

fonduer/parser/parser.py
```diff
@@ -127,22 +127,36 @@ def apply(self, x, **kwargs):
                 logger.error("Visual parsing failed: pdf_path is required")
             for _ in self.parse(document, text):
                 pass
+            if not self._valid_pdf(self.pdf_path, document.name):
+                logger.error(
```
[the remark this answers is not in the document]
lukehsiao: Good idea.

PR #106, opened 2018-08-20
move test data to server

SenWu:


SenWu:
Good catch! Contributors still can update their own test data (we assume they are small). Moving the current test data to the server only because they are large files.

PR #108, opened 2018-08-20
Fix something that looks like a bug

HiromuHota:
I'm not sure, but this looks like a bug to me.
@lukehsiao , please take a look and please add a unit test that fails on the current codebase.


lukehsiao:
Yes, this looks like a bug.

PR #110, opened 2018-08-21
Change import syntax to be more explicit

lukehsiao:
This refactor eliminates the host of imports that occurred at
`fonduer/__init__.py` and opts for more explicit import syntax to make it
clearer to the user which pipeline phases each import is associated
with. With this change, the public API of each pipeline phase is more
clearly defined in each pipeline's respective `__init__.py`.

In addition, this commit splits the old annotations.py file into two
separate files, one for the Annotator classes, and one of annotation
utilities.


lukehsiao:
Going to refactor our docs so that people know what can be imported from each package before merging.

PR #114, opened 2018-08-22
Reorganize supervision, learning, and util models

lukehsiao:


fonduer/supervision/models/__init__.py
```diff
@@ -1,3 +1,9 @@
-from fonduer.supervision.models.annotation import GoldLabel, GoldLabelKey
+from fonduer.supervision.models.label import (
+    GoldLabel,
+    GoldLabelKey,
+    Label,
+    LabelKey,
+    StableLabel,
```
lukehsiao: I don't actually know if StableLabel is used anywhere. Seems like it is not.

PR #123, opened 2018-08-29
Introduce mocks to the tests and decouple test_parser from db

HiromuHota:
Most of the current unit tests require a database access, but some of them should be decoupled from it to do **unit** test.
The patch in the first submission only modifies just one unit test.
If the first patch is acceptable, I can modify more.


tests/parser/test_parser.py
```diff
@@ -13,44 +13,55 @@
 from fonduer import Meta
 from fonduer.parser import Parser
 from fonduer.parser.models import Document, Sentence
+from fonduer.parser.parser import ParserUDF
 from fonduer.parser.preprocessors import HTMLDocPreprocessor
+from fonduer.parser.spacy_parser import Spacy
+from fonduer.utils.udf import UDF
 
 ATTRIBUTE = "parser_test"
 ATTRIBUTE = "parser_test"
```
lukehsiao: Can you help us delete this redundant line?

PR #127, opened 2018-08-30
Refactor featurizer and labeler to use sqlalchemy bulk inserts and support multiple relations

lukehsiao:
This PR accomplishes a few main goals:

1. The ability to access labels and features directly from candidates.
2. To use sqlalchemy features for bulk inserts rather than relying on hand-written sql queries.
3. Simplified API that allows users to more intuitively interact with candidates rather than sqlalchemy queries.


lukehsiao:
I'm seeing this
```
psycopg2.OperationalError: SSL error: decryption failed or bad record mac
```
when running on tutorials. Let me debug more before a review.

tests/e2e/test_e2e.py
```diff
@@ -265,22 +322,24 @@ def test_e2e(caplog):
...
-    )
+    labeler.update(split=0, lfs=[stg_temp_lfs_2], parallelism=PARALLEL)
+    assert session.query(Label).count() == 3346
+    assert session.query(LabelKey).count() == 13
+    L_train = labeler.get_label_matrices(train_cands)
+    assert L_train[0].shape == (3346, 13)
 
     gen_model = GenerativeModel(cardinalities=2)
-    gen_model.train(L_train, n_epochs=500, print_every=100)
+    gen_model.train(L_train[0], n_epochs=500, print_every=100)
```
SenWu: Can we comment on​ why it's [0]?

tests/e2e/test_e2e.py
```diff
@@ -166,18 +165,30 @@ def test_e2e(caplog):
...
+            )[0]
+        )
+        == 70
+    )
 
     # Candidate Extraction
     PartTemp = candidate_subclass("PartTemp", [Part, Temp])
 
-    candidate_extractor = CandidateExtractor([PartTemp], throttlers=[temp_throttler])
+    candidate_extractor = CandidateExtractor(
```
SenWu: Is there a way to hide session in mention_extractor, candidate_extractor, featurizer, and labeler?

tests/e2e/test_e2e.py
```diff
@@ -69,13 +71,10 @@ def test_e2e(caplog):
...
-    if num_docs != max_docs:
-        logger.info("Parsing...")
-        corpus_parser = Parser(
-            structural=True, lingual=True, visual=True, pdf_path=pdf_path
-        )
-        corpus_parser.apply(doc_preprocessor, parallelism=PARALLEL)
+    corpus_parser = Parser(
+        session, structural=True, lingual=True, visual=True, pdf_path=pdf_path
+    )
+    corpus_parser.apply(doc_preprocessor, parallelism=PARALLEL)
```
SenWu: Move parallelism to the constructor?

src/fonduer/utils/utils_udf.py
```diff
@@ -0,0 +1,158 @@
...
+    stmt = insert(table.__table__).values(records)
+    stmt = stmt.on_conflict_do_update(
+        constraint=table.__table__.primary_key,
+        set_={"keys": stmt.excluded.get("keys"), "values": stmt.excluded.get("values")},
+    )
+    session.execute(stmt)
+    session.commit()
+
+
+def get_sparse_matrix(session, key_table, cand_lists, key=None):
```
SenWu: Beside this, we probably need a function call get_keys which returns all the keys in the sparse matrix in order.

src/fonduer/utils/models/annotation.py
```diff
@@ -10,84 +11,55 @@
...
 class AnnotationKeyMixin(object):
-    """
-    Mixin class for defining annotation key tables. An AnnotationKey is the
-    unique name associated with a set of Annotations, corresponding e.g. to a
-    single labeling or feature function.  An AnnotationKey may have an
-    associated weight (Parameter) associated with it.
+    """Mixin class for defining annotation key tables.
+
+    An AnnotationKey is the unique name associated with a set of Annotations,
+    corresponding e.g. to a single labeling or feature function.
```
SenWu: labeling or feature function?

src/fonduer/supervision/models/label.py
```diff
@@ -1,4 +1,5 @@
 from sqlalchemy import Column, Integer, String
+from sqlalchemy.dialects import postgresql
```
SenWu: Do you think wrap up all postgresSQL specific functions to one place will be easier for user to extend for other databases?

src/fonduer/supervision/labeler.py
```diff
@@ -0,0 +1,237 @@
...
+    ):
+        """Call the LabelerUDF."""
+        if lfs is None:
+            raise ValueError("Please provide a list of labeling functions.")
+
+        if len(lfs) != len(self.candidate_classes):
+            raise ValueError("Please provide LFs for each candidate class.")
+
+        self.lfs = lfs
+        if docs:
```
SenWu: Can we add docstring to mention that doc has higher priority than split?

src/fonduer/supervision/labeler.py
```diff
@@ -0,0 +1,237 @@
...
+
+        for i in range(len(self.lfs)):
+            # Filter out the new/updated LFs
+            self.lfs[i] = [
+                lf
+                for lf in self.lfs[i]
+                if lf.__name__ not in [_.__name__ for _ in lfs[i]]
+            ]
+            # Then add them
+            self.lfs[i].extend(lfs[i])
```
SenWu: Will this rerun all old lfs?

src/fonduer/candidates/candidates.py
```diff
@@ -56,22 +59,62 @@ def apply(self, xs, split=0, **kwargs):
...
             ).filter(Candidate.split == split).delete()
 
-    def clear_all(self, session, split, **kwargs):
+    def clear_all(self, split, **kwargs):
         """Delete all Candidates from given split the database."""
         logger.info("Clearing ALL Candidates.")
-        session.query(Candidate).filter(Candidate.split == split).delete()
+        self.session.query(Candidate).filter(Candidate.split == split).delete()
+
+    def get_candidates(self, docs=None, split=0):
```
SenWu: It seems this is highly redundant with get_mentions. Can we reuse most of the code?

src/fonduer/supervision/labeler.py
lukehsiao: Yes, for the candidate which need to be updated (e.g. if you change the definition of several of the old LFs, we want to reevaluate them.)

src/fonduer/supervision/labeler.py
lukehsiao: 👍 good idea

src/fonduer/supervision/models/label.py
lukehsiao: This is not easy. We need a major refactor to be able to decouple things cleanly. If we can decouple fonduer and the database, switching to other databases and testing will both be easier. But, that change is definitely out of scope for this PR.

src/fonduer/utils/models/annotation.py
lukehsiao: Nice catch! Will fix.

src/fonduer/utils/utils_udf.py
lukehsiao: Ok, which classes should expose that function? Labeler and Featurizer?

tests/e2e/test_e2e.py
lukehsiao: Hmmm, this would restructure all of the UDF classes, but doesn't seem like a bad idea. It would not allow the user to change the parallelism between runs. For example, in our tests, we would need to create new objects to test single vs multithreaded, rather than just being able to change a parameter like we do now.

tests/e2e/test_e2e.py
lukehsiao: This seems tricky. That would basically embed the session into the UDF class. I imagine it's possible, but makes the database even more coupled than it is now.

tests/e2e/test_e2e.py
lukehsiao: 👍

tests/pytest.ini
```diff
@@ -1,6 +1,6 @@
 [pytest]
 addopts = -v -rsXx
-; addopts = -v -rsXx -s
+; addopts = -v -rsXx -s -x --pdb
```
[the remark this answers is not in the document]
lukehsiao: That line is commented out and isn't actually used. The `-x --pdb` is very useful for local debugging, along with the log_cli = true. If we don't want any of these comments for devs, we can remove both lines, but they are somewhat useful.

tests/e2e/test_e2e.py
```diff
@@ -265,22 +322,24 @@ def test_e2e(caplog):
         LF_temp_outside_table,
         LF_not_temp_relevant,
     ]
-
-    labeler = LabelAnnotator(PartTemp, lfs=stg_temp_lfs_2)
-    L_train = labeler.apply(
-        split=0, clear=False, update_keys=True, update_values=True, parallelism=PARALLEL
-    )
+    labeler.update(split=0, lfs=[stg_temp_lfs_2], parallelism=PARALLEL)
```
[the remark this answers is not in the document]
lukehsiao: (see above)

src/fonduer/candidates/candidates.py
lukehsiao: 👍 good idea

src/fonduer/supervision/labeler.py
lukehsiao: Yes, this will. In general the number of LFs is small so it seem like this overhead is small relatively small as well as it just running some several python functions and ultimately still is only a single upsert in the DB. If we do not want to rerun old LFs, I can refactor this.

tests/e2e/test_e2e.py
lukehsiao: Parallelism isn't required (in that it's a named argument with a default value of 1). I'll take a look at adding it to the constructors.

tests/pytest.ini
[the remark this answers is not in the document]
lukehsiao: Definitely.

src/fonduer/supervision/labeler.py
SenWu: I think we'd not rerun old/untouched/unchanged LFs so we don't need to refactor again.

src/fonduer/parser/parser.py
```diff
@@ -32,10 +32,10 @@ class Parser(UDFRunner):
     def __init__(
         self,
         session,
+        parallelism=None,
```
SenWu: Why not just set to 1?

src/fonduer/candidates/candidates.py
```diff
@@ -38,11 +39,13 @@ def __init__(
         self_relations=False,
         nested_relations=False,
         symmetric_relations=True,
+        parallelism=None,
```
SenWu: Why not just set to 1?

src/fonduer/parser/parser.py
```diff
@@ -32,10 +32,10 @@ class Parser(UDFRunner):
...
         structural=True,  # structural information
         blacklist=["style", "script"],  # ignore tag types, default: style, script
         flatten=["span", "br"],  # flatten tag types, default: span, br
-        flatten_delim="",
```
SenWu: what did this param do before?

src/fonduer/supervision/labeler.py
```diff
@@ -82,13 +89,7 @@ def apply(
             # Call apply on the specified docs for all splits
             split = ALL_SPLITS
             super(Labeler, self).apply(
-                docs,
-                split=split,
-                train=train,
-                update=update,
```
SenWu: why remove update?

tests/e2e/test_e2e.py
```diff
@@ -51,7 +51,7 @@
 DB = "e2e_test"
 
 
-@pytest.mark.skipif("CI" not in os.environ, reason="Only run e2e on Travis")
+#  @pytest.mark.skipif("CI" not in os.environ, reason="Only run e2e on Travis")
```
SenWu: Uncomment out this?

src/fonduer/supervision/labeler.py
lukehsiao: Now that we are using a upsert for everything, update doesn't really serve a purpose.

src/fonduer/candidates/candidates.py
lukehsiao: 1 is equivalent to None in the code.
```py
        if parallelism is None or parallelism < 2:
            self.apply_st(xs, clear=clear, **kwargs)
```
I'm fine either way. Want it to be `1`?

src/fonduer/parser/parser.py
lukehsiao: (see above)

src/fonduer/parser/parser.py
lukehsiao: Nothing from what I can tell. It's completely unused on `master`. Just stored for no reason.

lukehsiao: This was used back in v0.1.8 before we refactored the Parser when we implemented _flatten ourselves.

src/fonduer/candidates/candidates.py
SenWu: I think 1 is clear and can reduce the code.

src/fonduer/parser/parser.py
SenWu: 👍

Issue #104, opened 2018-08-17
Support more languages


HiromuHota:
**Is your feature request related to a problem? Please describe.**
My documents are written in Japanese, which is not supported by SpaCy hence not by Fonduer.

**Describe the solution you'd like**
According to [SpaCy](https://spacy.io/usage/models), tokenization of Japanese and other languages is alpha supported.
Please support these languages if tokenization is useful than nothing for Fonduer.

j-rausch:
We added support for Japanese spacy tokenization in #130. More languages that are alpha-supported by spacy could in principle be added in the future. See also our FAQ: https://fonduer.readthedocs.io/en/latest/user/faqs.html#how-can-i-use-use-fonduer-for-documents-in-languages-other-than-english

HiromuHota:
Thank you so much for your hard work!
Now that even alpha-supported languages can theoretically be added in the future, this issue can be closed.
If someone wants another language to be supported (by Fonduer), he/she can create another issue for that specific language.

PR #130, opened 2018-08-31
Multi language

j-rausch:
Added spacy parsing support for non-English languages. If (alpha) language support is available, the tokenizer will now default to the spacy implementation. Using the `lingual=True` parser argument will now activate the spacy NLP pipeline after sentence splitting/tokenization.

See #104.


docs/user/faqs.rst
```diff
@@ -61,3 +61,34 @@ testing on 14.04 or older.
 
 .. _see changelog: https://poppler.freedesktop.org/releases.html
 .. _install manually: https://poppler.freedesktop.org
+
+How can I use use Fonduer for documents in Languages other than English?
+--------------------------------------------------------------------------------
```
lukehsiao: Make this header line the same length as the header
```
How can I use use Fonduer for documents in Languages other than English?
------------------------------------------------------------------------
```

docs/user/faqs.rst
```diff
@@ -61,3 +61,34 @@ testing on 14.04 or older.
...
+
+    $ sudo apt-get install libmecab-dev
+    $ sudo apt-get install mecab mecab-ipadic-utf8
+
+For OSX::
+
+    $ brew install mecab
+    $ brew install mecab-ipadic
+
+Afterwards, you can use the ``make dev_extra`` command to install Fonduer with
```
lukehsiao: Only developers should be dealing with `make`ing the dev environment. Since this doc is for end users, they should be doing something like 

```
pip install fonduer[spacy_ja]
```

I'd also make it explicit in the following sentence that Japanese is the _only_ extra language integrated in Fonduer.

lukehsiao:
This PR also drops our coverage by 3.6%. It would be nice if we could try and add some tests to get some of that coverage back.

docs/user/faqs.rst
```diff
@@ -61,3 +61,34 @@ testing on 14.04 or older.
...
+
+For OS X::
+
+    $ brew install mecab
+    $ brew install mecab-ipadic
+
+Afterwards, you can use ``pip install fonduer[spacy_ja]`` to install Fonduer with
+Japanese language support. 
+If you would like to use other languages with spaCy alpha support, which are not yet integrated in Fonduer,
+feel free to open an Issue_.
```
lukehsiao: Sorry for nit: can you wrap these at 80-chars?

src/fonduer/parser/spacy_parser.py
```diff
@@ -94,7 +51,25 @@ class Spacy(object):
     def __init__(self, lang):
         self.logger = logging.getLogger(__name__)
         self.name = "spacy"
-        self.model = Spacy.load_lang_model(lang)
+        self.spacy_languages = ["en", "de", "es", "pt", "fr", "it", "nl", "xx"]
+        self.alpha_languages = {"ja": "Japanese"}
```
HiromuHota: `spacy_languages` and `alpha_languages` are a bit confusing because "spacy" and "alpha" are not counterparts.
Can you make it `languages` and `alpha_languages`, or `supported_languages` and `alpha_supported_languages`, or something like these?

PR #131, opened 2018-08-31
clean torchvision dependency

SenWu:


CHANGELOG.rst
```diff
@@ -54,6 +54,7 @@ Changed
 
 Removed
 ^^^^^^^
+* `@senwu`_: Remove torchvision from the dependency list.
```
lukehsiao: This actually doesn't need to be here. The changelog is what has changed from v0.2.3 to v0.3.0. This hasn't been released, so this line should be removed.

PR #133, opened 2018-09-01
Streaming input

j-rausch:
- Changed UDF input to a streaming queue in order to start parsing documents, as soon as the first documents are fed to the queue. In cases where there was is large number of input documents, parsing would otherwise be delayed.
- Added some small changes to HTML and Doc preprocessors:
    - HTML preprocessor: Scan optional directory to check for matching PDF files during initialization. 
    - Doc preprocessor: Changed the way the length of the DocPreprocessor is calculated to avoid wrong return values for cases in which `max_docs > number_of_html_files_in_folder

### DIFF
--- a/src/fonduer/parser/preprocessors/html_doc_preprocessor.py
+++ b/src/fonduer/parser/preprocessors/html_doc_preprocessor.py
@@ -13,7 +13,7 @@ class HTMLDocPreprocessor(DocPreprocessor):
     def __init__(self, *args, **kwargs):
         matching_pdf_folder = kwargs.pop("matching_pdf_folder", None)
         super(HTMLDocPreprocessor, self).__init__(*args, **kwargs)
-        if matching_pdf_folder is not None:
+        if matching_pdf_folder:
             pdf_files = [
                 x.lower()
                 for x in os.listdir(matching_pdf_folder)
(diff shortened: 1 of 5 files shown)


src/fonduer/parser/preprocessors/html_doc_preprocessor.py
```diff
@@ -10,6 +10,25 @@
 class HTMLDocPreprocessor(DocPreprocessor):
     """Simple parsing of files into html documents"""
 
+    def __init__(self, *args, **kwargs):
+        matching_pdf_folder = kwargs.pop("matching_pdf_folder", None)
+        super(HTMLDocPreprocessor, self).__init__(*args, **kwargs)
+        if matching_pdf_folder is not None:
```
lukehsiao: Just `if matching_pdf_folder`.

lukehsiao:
One more question from me here is how this is to be used by the user, and how to bridge this and the `Parser` itself.

For example, an `HTMLDocPreprocessor` is in theory, just to "preprocess" HTML documents. PDFs are only relevant if `visual=True` in the `Parser` object, which also has the pdf_path. 

@SenWu Any thoughts on how to make this API cleaner/more intuitive? Right now it looks like a user would need to pass the pdf_path to both the `HTMLDocPreprocessor` and the `Parser`.

j-rausch:
Good point. Right now, the parser is decoupled from the input (html) data generator, which seems like a good idea.
We could do something similar, and check/select the corresponding PDF file for visual linking outside of the parser. Afterwards, we could use, for example, ``ParserUDF.apply(self, (docA,textA), pdf=pdf_pathA **kwargs)``.

SenWu:
@lukehsiao I think we should keep the API simple and do not let the user provide arguments multiple times.

PR #136, opened 2018-09-05
Updating progress bar parallel

j-rausch:
I added a fix to the way the input queue is filled in parallel mode, such that the progress bar will continuously be updated during parsing again.


j-rausch:
I also modified the code such that the input queue filling should now be much faster. Instead of a single loop, we fill the queue with a pool of processes asynchronously.

PR #142, opened 2018-09-08
Changed to single worker UDF queue filling; Changed DocPreprocessor and subclasses yield objects

j-rausch:
1. Changed UDF input queue filling to one single worker process in order to properly handle input types which cannot be handled properly in parallel only via `__getitem__` calls.
2. `DocPreprocessor` and subclasses now yield `Document` objects instead of `(Document, Text)` tuples.

### DIFF
--- a/src/fonduer/parser/parser.py
+++ b/src/fonduer/parser/parser.py
@@ -151,9 +151,9 @@ class ParserUDF(UDF):
             self.pdf_path = pdf_path
             self.vizlink = VisualLinker()
 
-    def apply(self, x, **kwargs):
-        # The document is the Document model. Text is string representation.
-        document, text = x
+    def apply(self, document, **kwargs):
+        # The document is the Document model
+        text = document.text
         if self.visual:
             if not self.pdf_path:
                 warnings.warn(
@@ -418,7 +418,10 @@ class ParserUDF(UDF):
                     if attr.find("style") >= 0:
                         cur_style_index = index
                         break
-                styles = state["root"].find("head").find("style")
+                head = state["root"].find("head")
+                styles = None
+                if head is not None:
+                    styles = head.find("style")
                 if styles is not None:
                     for x in list(context_node.attrib.items()):
                         if x[0] == "class":
--- a/src/fonduer/parser/preprocessors/doc_preprocessor.py
+++ b/src/fonduer/parser/preprocessors/doc_preprocessor.py
@@ -26,28 +26,30 @@ class DocPreprocessor(object):
         """
         doc_count = 0
         for fp in self.all_files:
-            file_name = os.path.basename(fp)
-            if self._can_read(file_name):
-                for doc, text in self.parse_file(fp, file_name):
-                    yield doc, text
-                    doc_count += 1
-                    if doc_count >= self.max_docs:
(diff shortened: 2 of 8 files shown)


src/fonduer/parser/parser.py
```diff
@@ -418,7 +418,10 @@ def _parse_sentence(self, paragraph, node, state):
                     if attr.find("style") >= 0:
                         cur_style_index = index
                         break
-                styles = state["root"].find("head").find("style")
+                head = state["root"].find("head")
+                styles = None
+                if head is not None:
```
lukehsiao: just `if head`

src/fonduer/parser/parser.py
```diff
@@ -418,7 +418,10 @@ def _parse_sentence(self, paragraph, node, state):
...
+                    styles = head.find("style")
                 if styles is not None:
```
lukehsiao: and `if styles`

src/fonduer/parser/preprocessors/doc_preprocessor.py
```diff
@@ -26,29 +26,31 @@ def generate(self):
...
-            for doc, text in self.parse_file(fp, file_name):
-                yield doc, text
+        return [doc for doc in self.get_docs_for_path(fp)]
 
     def __len__(self):
-        """Provide a len attribute based on max_docs and number of files in folder."""
-        num_docs = min(len(self.all_files), self.max_docs)
-        return num_docs
+        raise NotImplementedError(
+            "One generec file can yield more than one Document objects, "
```
lukehsiao: `generec` --> `generic`

`one Document objects` --> `one Document object`

src/fonduer/parser/preprocessors/doc_preprocessor.py
```diff
@@ -26,29 +26,31 @@ def generate(self):
...
+                    return
 
-    def parse_by_index(self, index):
+    def __getitem__(self, index):
         fp = self.all_files[index]
-        file_name = os.path.basename(fp)
-        if self._can_read(file_name):
-            for doc, text in self.parse_file(fp, file_name):
-                yield doc, text
+        return [doc for doc in self.get_docs_for_path(fp)]
```
lukehsiao: Is this correct? `__getitem__` returns all the documents, rather than one?

j-rausch: Thanks! This is not used anymore and was wrong in its current form.

src/fonduer/utils/udf.py
```diff
@@ -57,17 +51,20 @@ def apply(self, xs, clear=True, parallelism=None, progress_bar=True, **kwargs):
...
-            self.pb = tqdm(total=len(xs))
+            if hasattr(doc_loader, "__len__"):
+                self.pb = tqdm(total=len(doc_loader))
+            else:
+                self.logger.error("Could not determine size of progress bar")
 
         # Use the parallelism of the class if none is provided to apply
         parallelism = parallelism if parallelism else self.parallelism
-
+        print("parallelism: {}".format(parallelism))
```
SenWu: Please use logging instead of print.

PR #145, opened 2018-09-11
Added n_min argument to the ngram class

j-rausch:
I added an `n_min` argument to the `Ngrams` class. This can be useful in cases, where we look for mentions with at least two tokens, e.g. names of persons (first name + last name).


SenWu:
Can we add a test for this?

lukehsiao:
Sorry, one ask to change. If possible, can we add a docstring, and can you update the CHANGELOG.rst to indicated that we added this new parameter?

PR #150, opened 2018-09-13
fix _get_node performance issue by using cache

### LINKED ISSUE #139: Matchers using data model utilities based on _get_node() are too slow
I was generating candidates from a document where I pretty much needed all the **mentions of the bold text in the first column in the tables:**

lambda function matcher initialisation:
```python
def get_style_attr(html_attr):
    """This function parse a style string and returns a dict of styles.
        eg: 'style=padding-left: 5pt;padding-right: 66pt;font-weight: normal;'
        returns => {
            'padding-left': '5pt',
            'padding-right': '66pt',
            'font-weight': 'normal'
        }
    """
    for attr in html_attr:
        if attr.startswith('style'):
            beg = attr.find('style=') + 6
            return {style.strip().split(':')[0].strip():style.strip().split(':')[1].strip()\
                        for style in attr[beg:].strip().split(';') if len(style.strip()) > 0}

def name_conditions(span):
    if get_parent_tag(span) != 'td':
        return False
    style_attr = get_style_attr(get_attributes(span))
    return style_attr is not None and 'font-weight' in style_attr.keys() and style_attr['font-weight'] == 'bold'

name_lambda_matcher = LambdaFunctionMatcher(func=name_conditions)
```
candidate extractor:

```python
name_label_extractor = CandidateExtractor(name_label, 
                        [name_ngrams, label_ngrams], 
                        [name_lambda_matcher, label_matcher],                              
                        candidate_filter = name_label_filter)

name_label_extractor.apply(docs, parallelism=4)
```
**I am running this on a document with 3620 phrases.**

Is this the expected time for this matcher..? or do you see anything here that is causing this..?

SenWu:
Closes #139

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -74,6 +74,7 @@ Removed
 
 Fixed
 ^^^^^
+* `@senwu`_: Speed-up of ``_get_node`` using caching.
 * `@HiromuHota`_: Fixed bug with Ngram splitting and empty TemporarySpans.
   (`#108 <https://github.com/HazyResearch/fonduer/pull/108>`_,
   `#112 <https://github.com/HazyResearch/fonduer/pull/112>`_)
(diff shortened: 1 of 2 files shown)


lukehsiao:
Resolved using LRU caching in #152.

PR #154, opened 2018-09-18
Add supporting functions for incrementally adding docs

lukehsiao:
The purpose of this PR is to allow users to incrementally add new documents, without rerunning the whole pipeline.

### Example Workflow

Assuming you already parsed a set of Documents, A, and wanted to add a new set of Documents, B, a user could now run something like the following, in the context of the [e2e test](https://github.com/HazyResearch/fonduer/blob/master/tests/e2e/test_e2e.py).

```py
    docs_path = "new_data/html/"
    pdf_path = "new_data/pdf/"

    doc_preprocessor = HTMLDocPreprocessor(docs_path, max_docs=max_docs)

    corpus_parser.apply(doc_preprocessor, pdf_path=pdf_path, clear=False)

    new_docs = corpus_parser.get_last_documents()

    # Get mentions from just the new docs
    mention_extractor.apply(new_docs, parallelism=PARALLEL, clear=False)

    # Just run candidate extraction and assign to split 0
    candidate_extractor.apply(new_docs, split=0, parallelism=PARALLEL, clear=False)

    # Update features
    featurizer.update(new_docs, parallelism=PARALLEL)

    # Update Labels
    labeler.update(new_docs, lfs=[stg_temp_lfs], parallelism=PARALLEL)
```

### DIFF
--- a/src/fonduer/features/featurizer.py
+++ b/src/fonduer/features/featurizer.py
@@ -36,6 +36,31 @@ class Featurizer(UDFRunner):
         )
         self.candidate_classes = candidate_classes
 
+    def update(self, docs=None, split=0, parallelism=None, progress_bar=True):
+        """Update the labels of the specified candidates based on the provided LFs.
+
+        :param docs: If provided, apply features to all the candidates in these
+            documents.
+        :param split: If docs is None, apply features to the candidates in this
+            particular split.
+        :type split: int
+        :param parallelism: How many threads to use for extraction. This will
+            override the parallelism value used to initialize the Featurizer if
+            it is provided.
+        :type parallelism: int
+        :param progress_bar: Whether or not to display a progress bar. The
+            progress bar is measured per document.
+        :type progress_bar: bool
+        """
+        self.apply(
+            docs=docs,
+            split=split,
+            train=True,
+            clear=False,
+            parallelism=parallelism,
+            progress_bar=progress_bar,
+        )
+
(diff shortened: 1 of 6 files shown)


src/fonduer/features/featurizer.py
```diff
@@ -36,6 +36,31 @@ def __init__(self, session, candidate_classes, parallelism=1):
         )
         self.candidate_classes = candidate_classes
 
+    def update(self, docs=None, split=0, parallelism=None, progress_bar=True):
+        """Update the labels of the specified candidates based on the provided LFs.
```
SenWu: LF?

tests/parser/test_parser.py
```diff
@@ -16,7 +16,7 @@ def get_parser_udf(
     language="en",
     lingual=True,  # lingual information
     strip=True,
-    replacements=[(u"[\u2010\u2011\u2012\u2013\u2014\u2212\uf02d]", "-")],
+    replacements=[(u"[\u2010\u2011\u2012\u2013\u2014\u2212]", "-")],
```
SenWu: What's this?

lukehsiao: Removing the invalid unicode.

src/fonduer/features/featurizer.py
lukehsiao: Nice catch!

PR #158, opened 2018-09-19
Skip documents if parse errors occur

j-rausch:
This changes the `parser.apply` function such that the parser only yields sentences, if no errors occurred during the whole parsing process.
It also adds a parser test case faulty test document.


tests/data/faulty_html/Bill_Gates.html
```diff
@@ -0,0 +1,310 @@
+<!DOCTYPE html>
```
lukehsiao: Is it possible to makes these example docs minimal (just to reduce the unnecessary lines of code).

PR #167, opened 2018-10-11
[WIP] Fix annotation key tables to support multiple relations

lukehsiao:


tests/e2e/test_e2e.py
```diff
@@ -373,40 +388,53 @@ def test_e2e(caplog):
...
     featurizer.apply(split=1, train=True, parallelism=PARALLEL)
-    assert session.query(Feature).count() == 72
-    assert session.query(FeatureKey).count() == 716
+    assert session.query(Feature).count() == 225
+    assert session.query(FeatureKey).count() == 1179
 
     # Test Dropping FeatureKey
     featurizer.drop_keys(["DDL_e1_W_LEFT_POS_3_[NFP NN NFP]"])
-    assert session.query(FeatureKey).count() == 715
+    assert session.query(FeatureKey).count() == 1178
```
SenWu: Can you explain the tests here? In case other people want to understand it.

CHANGELOG.rst
```diff
@@ -1,6 +1,16 @@
 [Unreleased]
 ------------
 
+Fixed
+^^^^^
+* `@lukehsiao`_: The features/labels/gold_label key tables were not properly
```
SenWu: Update this part with solving potential deadlock issue.

lukehsiao: I was going to do that, but realized that the deadlock issue was only there because of the changes made in this PR. The deadlock issue shouldn't have happened when it was just on_conflict_do_nothing.

PR #169, opened 2018-10-17
remove psycopg2

SenWu:


CHANGELOG.rst
```diff
@@ -1,5 +1,9 @@
 [Unreleased]
 ------------
+Changed
```
lukehsiao: Make this "Removed"

```suggestion
Removed
```

src/fonduer/meta.py
```diff
@@ -16,49 +15,35 @@ def new_sessionmaker():
...
             isolation_level="AUTOCOMMIT",
         )
-    else:
+        engine.connect()
+    except Exception:
         raise ValueError(
             "Meta variables have not been initialized with "
             "a valid postgres connection string."
+            "Use the form: "
+            "postgresql://<user>:<pw>@<host>:<port>/<database_name>"
```
lukehsiao: So this means we need to change `postgres` to `postgresql` in all the tutorials as well, right?

SenWu: yes

PR #175, opened 2018-11-01
SparseLSTM

SenWu:


CHANGELOG.rst
```diff
@@ -1,6 +1,11 @@
 [Unreleased]
 ------------
 
+Changed
```
lukehsiao: This should be `Added`. It would also be nice to add an example or description of when to use it in a note box:

```
.. note::
   content
```

PR #179, opened 2018-11-15
Improve error messages on meta connection

### LINKED ISSUE #178: Improve error message on Meta failure
**Describe the bug**
The FAQs include one for handling "fe_sendautho no password supplied", which typically occurs on a machine where postgres was freshly installed. However, in Fonduer v0.3.5, this error is caught and presented as a more generic error:

https://github.com/HazyResearch/fonduer/blob/50b9c322a2be79b70d22100977f80778978517c7/src/fonduer/meta.py#L26-L32

**Expected behavior**
We should provide more descriptive errors. If the Exception is caused by the fe_send_auth issue, the user should be directly pointed to the FAQ so they can fix it. If it is not the fe_send_auth issue, we can raise the more generic ValueError.

 - OS: e.g. Ubuntu 16.04
 - PostgreSQL Version: 11.1
 - Fonduer Version: 0.3.5

lukehsiao:
Closes #178.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,7 +29,7 @@ notifications:
 
 before_install:
 - if [ "$TRAVIS_OS_NAME" = "linux" ]; then OS=Linux-x86_64; else OS=MacOSX-x86_64; fi
-- if [ "$TRAVIS_OS_NAME" = "linux" ]; then sudo apt-get install -y libmecab-dev mecab mecab-ipadic-utf8; fi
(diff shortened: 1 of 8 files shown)


.travis.yml
```diff
@@ -71,7 +71,22 @@ install:
...
+sudo checkinstall make install
+    wget https://poppler.freedesktop.org/poppler-0.62.0.tar.xz
+    tar -xf poppler-0.62.0.tar.xz
+    cd poppler-0.62.0
+    mkdir build
+    cd build
+    cmake ..
+    sudo checkinstall make install > /dev/null 2>&1
+    cd ../..
+    rm -rf poppler-0.53.0
```
SenWu: 0.62.0?

PR #182, opened 2018-11-26
Improve data model

SenWu:
1. add support to parse multiple sections in parser.
2. fix webpage context.
3. add name column for each context in data model.


lukehsiao:
Please update `_version.py` to `0.4.0` also.

PR #189, opened 2018-12-06
Add soft version pinning to dependencies

### LINKED ISSUE #187: Loosely pin dependencies in setup.py
We want to avoid the issue we've seen where one of our dependencies update and consequently breaks `fonduer`. 

I'm imagining something like

> If you're releasing a library on PyPI, you should declare whatever dependencies you know about, but not pin to a specific version. For example, if you know you need >= 1.2, but 1.4 is broken, then you can write something like somepkg >= 1.2, != 1.4. If one of the things you know is that somepkg follows SemVer, then you can add a < 2.

Basically, something similar to [caret requirements](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#caret-requirements) that are default in Rust (but we'll mainly care about major versions).

Having ranges specified will be better than an exact pin, so we still benefit from updates in the dependencies, but can manually handle the API-breaking major changes.

lukehsiao:
Closes #187.


setup.py
```diff
@@ -14,23 +14,23 @@
...
-        "wand",
+        "bs4>=0.0.1, <0.1.0",
+        "editdistance>=0.5.2,<0.6.0",
+        "lxml>=4.2.5, <5.0.0",
+        "nltk>=3.3, <4.0",
+        "numpy>=1.11,<2.0",
+        "pandas>=0.23.4,<0.24.0",
+        "pyyaml>=3.13,<4.0",
+        "scipy>=1.1.0, <2.0.0",
+        "snorkel-metal>=0.3.2, <0.4.0",
```
SenWu: Can we change it to 0.3.3?

lukehsiao: Yes, definitely.

PR #191, opened 2018-12-08
fix non-deterministic issue from get_candidates and get_mentions

SenWu:
fix non-deterministic issue from get_candidates and get_mentions caused by parallel candidate/mention generation


SenWu:
No, I don't think it will affect performance too much.

Issue #186, opened 2018-12-03
`get_row_ngrams` and `get_col_ngrams` return ngrams for mentions not inside tables


j-rausch:
**Describe the bug**
`get_row_ngrams` and `get_col_ngrams` in https://github.com/HazyResearch/fonduer/blob/master/src/fonduer/utils/data_model_utils/tabular.py utilize `_get_axis_ngrams`.

If the given mention is not inside a table at all, the two functions could still return ngrams, because no `is_tabular()` check is performed before `get_sentence_ngrams` is called in https://github.com/HazyResearch/fonduer/blob/master/src/fonduer/utils/data_model_utils/tabular.py#L428.

 **Expected behavior**
I would expect `get_col_ngrams` and `get_row_ngrams` to return `None` or an empty list, if the mention is not inside a table (and not in a row/column as a consequence).

SenWu:
@j-rausch Will you submit a PR to fix it? Thanks!

j-rausch:
Sure, just wanted to make sure you guys agree on the desired behavior.

PR #194, opened 2018-12-10
Modified get_row_ngrams and get_col_ngrams to return None for non-table mentions

j-rausch:
I changed the behavior of `get_row_ngrams` and `get_col_ngrams` to return `None`, if a non-table mention is passed (see #186).
I chose `None`, as `[]` can be a valid return value at the moment, e.g. if all other cells in the row of a  `Mention` are empty.

Note that the current implementation of both functions also work on _candidates_, which will result in a list of ngrams being returned, even if only one of them is inside a table.

### DIFF
--- a/src/fonduer/utils/data_model_utils/tabular.py
+++ b/src/fonduer/utils/data_model_utils/tabular.py
@@ -135,6 +135,20 @@ def get_min_col_num(mention):
         return None
 
 
+def get_min_row_num(mention):
+    """Return the lowest row number that a Mention occupies.
+
+    :param mention: The Mention to evaluate. If a candidate is given, default
+        to its first Mention.
+    :rtype: integer or None
+    """
+    span = _to_span(mention)
+    if span.sentence.is_tabular():
+        return span.sentence.cell.row_start
+    else:
+        return None
+
+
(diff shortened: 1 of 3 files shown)


src/fonduer/utils/data_model_utils/tabular.py
```diff
@@ -425,6 +439,11 @@ def _get_axis_ngrams(
     mention, axis, attrib="words", n_min=1, n_max=1, spread=[0, 0], lower=True
 ):
     span = _to_span(mention)
+
+    if not span.sentence.is_tabular():
+        yield None
+        return
```
lukehsiao: Looking at your test case, it looks like this should either yeild ngrams, or a single `[None]`, is that correct? Is

```py
yield None
return
```

different than

```py
return None
```

?

j-rausch: 1) Yes, that's the intended behavior. 
2) Also yes, replacing both lines with `return None` would result in `list(get_row_ngrams(mention))` being an empty list `[]` instead of `[None]` for a non-tabular mention.

lukehsiao: Got it.

lukehsiao:
Before we merge it, can you update the CHANGELOG with this fix?

PR #196, opened 2018-12-14
Reattach doc with the current session at MentionExtractorUDF#apply to avoid doing so at each MentionSpace

HiromuHota:


src/fonduer/candidates/mentions.py
```diff
@@ -535,22 +526,22 @@ def __init__(self, mention_classes, mention_spaces, matchers, **kwargs):
 
         super(MentionExtractorUDF, self).__init__(**kwargs)
 
-    def apply(self, context, clear, **kwargs):
+    def apply(self, doc, clear, **kwargs):
         """Extract mentions from the given Context.
 
         :param context: A document to process.
```
lukehsiao: Can you make sure these changed parameter names are also updated in the docstrings?

src/fonduer/candidates/mentions.py
```diff
@@ -132,7 +132,7 @@ def __init__(self, n_min=1, n_max=5, split_tokens=[]):
         """
         Ngrams.__init__(self, n_min=n_min, n_max=n_max, split_tokens=split_tokens)
 
-    def apply(self, session, doc):
+    def apply(self, doc):
         """Generate MentionNgrams from a Document by parsing all of its Sentences.
 
         :param session: The database session
```
lukehsiao: Session can be removed from the docstring if it's removed from the parameter list.

PR #199, opened 2018-12-20
fix parser with multi-doc per file

SenWu:
In stead of using the number of files as `__len__` in `TSVDocPreprocessor` and `CSVDocPreprocessor`, in this fix it uses the actual number of rows as `__len__`. This fix will have the proper behavior with multiprocessing.


tests/parser/test_parser.py
```diff
@@ -627,14 +628,15 @@ def test_tsv_doc_preprocessor(caplog):
 
     # Test tsv document
     docs_path = "tests/data/various_format/tsv_format.tsv"
-    preprocessor = TSVDocPreprocessor(docs_path, max_docs=1)
```
lukehsiao: We should still have a test which explores the code path of max_docs actually limiting the number of documents parsed.

PR #202, opened 2018-12-26
Update code to use Python 3.6+'s f-strings

SenWu:


CHANGELOG.rst
```diff
@@ -45,6 +45,7 @@ Added
 
 Fixed
 ^^^^^
+* `@senwu`_: Update code to use Python 3.6+'s f-strings
```
lukehsiao: Let's put this under "Changed"

src/fonduer/learning/disc_learning.py
```diff
@@ -248,19 +249,18 @@ def train(
...
+                    f"Average loss={torch.stack(iteration_losses).mean():.6f}"
                 )
                 if X_dev is not None:
                     scores = self.score(
                         _X_dev, Y_dev, batch_size=self.settings["batch_size"]
                     )
                     score = scores if self.cardinality > 2 else scores[-1]
                     score_label = "Acc." if self.cardinality > 2 else "F1"
-                    msg += "\tDev {0}={1:.2f}".format(score_label, 100.0 * score)
+                    msg += f"\tDev {score_label}={100.0 * score:.2f}"
```
lukehsiao: Should double check that the order of operations here results in the correct formatting.

i.e., does this correctly execute as 
```py
(100.0 * score):.2f
```
?

src/fonduer/learning/disc_learning.py
```diff
@@ -277,9 +277,7 @@ def train(
 
         # Conclude training
         if verbose:
-            self.logger.info(
-                "[{0}] Training done ({1:.2f}s)".format(self.name, time() - st)
-            )
+            self.logger.info(f"[{self.name}] Training done ({time() - st:.2f}s)")
```
lukehsiao: Same here

src/fonduer/candidates/models/candidate.py
```diff
@@ -53,9 +53,11 @@ def __getitem__(self, key):
         return self.get_mentions()[key]
 
     def __repr__(self):
-        return "%s(%s)" % (
-            self.__class__.__name__,
-            ", ".join(map(str, self.get_mentions())),
+        return (
+            f"{self.__class__.__name__}"
+            f"("
+            f"{', '.join(map(str, self.get_mentions()))}"
```
lukehsiao: Should double check that this work correctly.

src/fonduer/parser/spacy_parser.py
```diff
@@ -297,16 +294,16 @@ def split_sentences(self, document, text):
             # temporary increase character limit of spacy
             # 'Probably save' according to spacy, as no parser or NER is used
             previous_max_length = self.model.max_length
-            self.model.max_length = 100000000
+            self.model.max_length = 100_000_000
```
lukehsiao: This is a nice readability improvement.

lukehsiao:
Let's also address the warnings that we can:

https://travis-ci.org/HazyResearch/fonduer/jobs/472224747

For example, the importing syntax from `collections` will be deprecated in Python 3.8.

src/fonduer/learning/disc_learning.py
lukehsiao: Checked, this works.

src/fonduer/candidates/models/candidate.py
lukehsiao: Confirmed that this works.

Issue #193, opened 2018-12-10
Upgrade to PyTorch 1.0


SenWu:
PyTorch 1.0 is out!

PR #206, opened 2019-01-08
Specify pytorch version as 0.4.1.post2 to safeguard from HazyResearch/metal#101

HiromuHota:
The PyTorch (0.4.1) for 3.7 was compiled without NumPy support (pytorch/pytorch#9903, HazyResearch/metal#101).
If Python 3.7 is used, the PyTorch should be 0.4.1.post2 instead of 0.4.1 (until Fonduer becomes compatible with PyTorch 1.0.0).
The current requirement `torch>=0.4.1, <0.5.0` is satisfied when torch==0.4.1 is installed already.
As a safeguard from using PyTorch of 0.4.1, the requirement should be specific.


setup.py
```diff
@@ -25,7 +25,7 @@
         "snorkel-metal>=0.3.3, <0.4.0",
         "spacy>=2.0.12, <3.0.0",
         "sqlalchemy[postgresql_psycopg2binary]>=1.2.12, <2.0.0",
-        "torch>=0.4.1, <0.5.0",
+        "torch==0.4.1.post2",
```
lukehsiao:
```suggestion
        "torch>=0.4.1.post2, <0.5.0",
```
I believe we should still be able to have an upper limit on the dependency, like this.

HiromuHota:
According to the travis result, 0.4.1.post2 is only available for Python 3.7 on Linux.
@lukehsiao close this without merging it unless you have a good solution.

lukehsiao:
@HiromuHota: This approach may work, right?

https://stackoverflow.com/questions/21082091/install-requires-based-on-python-version:
```py
install_requires=[
    ...
    "torch>=0.4.1.post2, <0.5.0; python_version >= 3.7",
    "torch>=0.4.1, <0.5.0; python_version < 3.7",
    ...
]
```

lukehsiao:
Later on, we should just upgrade to pytorch 1.0 (See #193).

PR #207, opened 2019-01-10
fix max/min in visualizer.get_box

KenSugimoto:
The bounding box value increases from left to right and from top to bottom, so I think the left bounding box value of the span should be "min" of words and the bottom should be "max" of words.


HiromuHota:
The test failed and said
> CHANGELOG.rst:9:Unknown target name: "@kensugimoto".

@lukehsiao Do you know how to fix this?

lukehsiao:
Yes, just add the link for that name at the very bottom of the file. You'll notice that's where all the other name links are put as well.

KenSugimoto:
Thanks for answering. I corrected it and pushed it again.

PR #209, opened 2019-01-27
update to pytorch 1.0

### LINKED ISSUE #193: Upgrade to PyTorch 1.0

SenWu:
Closes #193.


setup.py
```diff
@@ -25,8 +25,7 @@
         "snorkel-metal>=0.3.3, <0.4.0",
         "spacy>=2.0.12, <3.0.0",
         "sqlalchemy[postgresql_psycopg2binary]>=1.2.12, <2.0.0",
-        'torch>=0.4.1, <0.5.0;python_version<"3.7" or platform_system=="Darwin"',
-        'torch>=0.4.1.post2, <0.5.0;python_version>="3.7" and platform_system=="Linux"',
+        "torch>=1.0, <1.1",
```
lukehsiao: With `torch` reaching 1.0, any api-breaking changes should require a major version bump, so we should be able to use the following:

```suggestion
        "torch>=1.0, <2.0",
```
We were only using the minor version for packages still in version 0.x, since minor verion updates indicate API changes.

Issue #213, opened 2019-02-06
sqlalchemy: ObjectDeletedError


lukehsiao:
**Describe the bug**
During Candidate Extraction, some of the UDFs crash with a `sqlalchemy.orm.exc.ObjectDeletedError` on impllicit spans.

**To Reproduce**
Unfortunately, this requires a large dataset to reproduce. I will update this with a minimal example if I can find one. Otherwise, I've currently seeing it on the full transistor dataset when extracting candidates as follows

```py
from fonduer.candidates import CandidateExtractor

candidate_extractor = CandidateExtractor(
    session,
    [PartStgTempMin, PartStgTempMax, PartPolarity, PartCeVMax],
    throttlers=[temp_throttler, temp_throttler, polarity_throttler, ce_v_max_throttler],
)

for i, docs in enumerate([train_docs, dev_docs, test_docs]):
        candidate_extractor.apply(docs, split=i, parallelism=PARALLEL)
        logger.info(
            f"PartStgTempMin in split={i}: "
            f"{session.query(PartStgTempMin).filter(PartStgTempMin.split == i).count()}"
        )
        logger.info(
            f"PartStgTempMax in split={i}: "
            f"{session.query(PartStgTempMax).filter(PartStgTempMax.split == i).count()}"
        )
        logger.info(
            f"PartPolarity in split={i}: "
            f"{session.query(PartPolarity).filter(PartPolarity.split == i).count()}"
        )
        logger.info(
            f"PartCeVMax in split={i}: "
            f"{session.query(PartCeVMax).filter(PartCeVMax.split == i).count()}"
        )


train_cands = candidate_extractor.get_candidates(split = 0)
dev_cands = candidate_extractor.get_candidates(split = 1)
test_cands = candidate_extractor.get_candidates(split = 2)

logger.info(f"Total train candidate: {len(train_cands[0])}")
logger.info(f"Total dev candidate: {len(dev_cands[0])}")
logger.info(f"Total test candidate: {len(test_cands[0])}")
```

**Expected behavior**
No errors. No UDFs crashing.

**Error Logs/Screenshots**
```
Process CandidateExtractorUDF-46:
Traceback (most recent call last):
  File "/usr/lib/python3.6/multiprocessing/process.py", line 258, in _bootstrap
    self.run()
  File "/lfs/raiders10/hdd/lwhsiao/repos/fonduer/src/fonduer/utils/udf.py", line 194, in run
[...]

PR #214, opened 2019-02-07
fix: synchronize session on all deletes

lukehsiao:
Note that this does have some performance implications [[1]](https://docs.sqlalchemy.org/en/latest/orm/query.html):

> The 'fetch' strategy results in an additional SELECT statement emitted and will significantly reduce performance.

But some overhead in performance when running `clear` and `clear_all()` seems worth it if it eliminates the sqlalchemy error.

See #213.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -12,6 +12,8 @@ Changed
 
 Fixed
 ^^^^^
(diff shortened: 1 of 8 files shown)


SenWu:
Does this fix the #213 ?

lukehsiao:
Yes, it does for me.

SenWu:
👍

Issue #227, opened 2019-03-22
Not compatible with spaCy v2.1.0


HiromuHota:
**Describe the bug**

Fonduer is not compatible with spaCy v2.1.0

**To Reproduce**
1. Install spaCy v2.1.0
2. Run tests

**Expected behavior**

Be compatible with spaCy v2.1.0 and higher

**Error Logs/Screenshots**

```
jovyan@1a70f98cfba1:~/fonduer$ pip list | grep spacy
spacy                    2.1.2       
jovyan@1a70f98cfba1:~/fonduer$ pytest tests/parser/test_parser.py::test_parse_md_details
====================================================================================== test session starts ======================================================================================
platform linux -- Python 3.7.1, pytest-4.3.1, py-1.8.0, pluggy-0.9.0 -- /opt/conda/bin/python
cachedir: .pytest_cache
rootdir: /home/jovyan/fonduer/tests, inifile: pytest.ini
collected 1 item                                                                                                                                                                                

tests/parser/test_parser.py::test_parse_md_details FAILED                                                                                                                                 [100%]

=========================================================================================== FAILURES ============================================================================================
_____________________________________________________________________________________ test_parse_md_details _____________________________________________________________________________________

caplog = <_pytest.logging.LogCaptureFixture object at 0x7f4728ad9e48>

    def test_parse_md_details(caplog):
        """Test the parser with the md document."""
        caplog.set_level(logging.INFO)
        logger = logging.getLogger(__name__)
    
        docs_path = "tests/data/html_simple/md.html"
        pdf_path = "tests/data/pdf_simple/md.pdf"
    
        # Preprocessor for the Docs
        preprocessor = HTMLDocPreprocessor(docs_path)
        doc = next(preprocessor._parse_file(docs_path, "md"))
    
        # Check that doc has a name
        assert doc.name == "md"
    
        # Check that doc does not have any of these
        assert len(doc.figures) == 0
        assert len(doc.tables) == 0
        assert len(doc.cells) == 0
        assert len(doc.sentences) == 0
    
        # Create an Parser and parse the md document
        parser_udf = get_parser_udf(
            structural=True,
            tabular=True,
            lingual=True,
            visual=True,
            pdf_path=pdf_path,
            language="en",
        )
>       for _ in parser_udf.apply(doc):

tests/parser/test_parser.py:80: 
[...]

PR #228, opened 2019-03-22
Make sure Fonduer uses spaCy v2.0.X until #227 gets resolved.

HiromuHota:


lukehsiao:
This will be superseded by #230.

Issue #231, opened 2019-03-25
For Japanese, Mecab should use unidic instead of ipadic


HiromuHota:
**Describe the bug**

Mecab is a text segmentation library for Japanese.
There are a few dictionaries for Mecab: ipadic, unidic, etc.
How a text is segmented depends on the dictionary in use.
spaCy assumes unidic is used (https://spacy.io/usage/models#languages).
However, [FAQ](https://fonduer.readthedocs.io/en/latest/user/faqs.html#how-can-i-use-use-fonduer-for-documents-in-languages-other-than-english) suggests to install Mecab with ipadic.

I'll make a PR for this.

PR #232, opened 2019-03-26
Use unidic instead of ipadic for Japanese (#231)

HiromuHota:


HiromuHota:
This is actually a bug fix rather than enhancement.

SenWu:
It failed on Mac tests, can you double check there? Thanks! @HiromuHota

HiromuHota:
I'll insert a sanity-check about mecab configuration to `.travis.yml`. Please bear with me for a while.

HiromuHota:
On mac, the sanity check command: `mecab -D` returns `param.cpp(69) [ifs] no such file or directory: /usr/local/lib/mecab/dic/ipadic/dicrc`

PR #244, opened 2019-04-12
Xenial comes with Poppler v0.41.0, which is greater than v0.36.0

HiromuHota:
This simplifies and speeds up the Travis CI, which currently spends roughly 140s on Poppler compile/install.

The doc says:
> Fonduer requires poppler-utils to be version 0.36.0 or above.

So this should work. Otherwise, the doc needs changing.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,7 +31,6 @@ notifications:
 before_install:
 - if [ "$TRAVIS_OS_NAME" = "linux" ]; then OS=Linux-x86_64; else OS=MacOSX-x86_64; fi
 - if [ "$TRAVIS_OS_NAME" = "linux" ]; then sudo apt-get install -q -y libmecab-dev swig mecab unidic-mecab; fi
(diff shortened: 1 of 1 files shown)


lukehsiao:
While this "works", our specific test cases require a newer version of poppler. If this was bionic, we wouldn't have to install anything, but xenial's version is still a bit too old to pass our tests specifically.

HiromuHota:
I would say Fonduer with Poppler v0.41.0 "runs" but does not work as expected on a certain condition.
I figured out that Poppler had an issue in splitting words in pdfs produced by Chrome (https://bugs.freedesktop.org/show_bug.cgi?id=97399), which has been fixed at v0.48.0.
For this particular pdf, diseases.pdf, was actually created by Chrome hence affected by this issue.
I tested locally and confirmed that v0.47.0 did not split words and v0.48.0 did split them.
Thus, Poppler requires to be v0.48.0 or later to produce an expected result.

This PR aims at simplifying .travis.yml and speeding up the CI process.
If Poppler v0.75.0 installed by Conda runs and works as expected, please merge this.

PR #245, opened 2019-04-12
Enable pip cache even when language: generic

HiromuHota:
According to [TravisCI](https://docs.travis-ci.com/user/caching/#caching-directories-bundler-dependencies), .travis.yml should use 
```
language: python
cache: pip
```
but the current .travis.yml uses `language: generic` and travis job log does not say anything about pip cache.

Hoping to re-enable pip caching, this patch specifies the cache directory manually instead of using caching shortcut (e.g., `cache: pip`).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,10 @@
 language: generic
 dist: xenial
-cache: pip
+cache:
+  directories:
+  - $HOME/.cache/pip
+before_cache:
+  - rm -f $HOME/.cache/pip/log/debug.log
 
 
 matrix:


HiromuHota:
The step that can benefit most from the pip caching would be `make -s dev_extra`.
At https://travis-ci.org/HazyResearch/fonduer/jobs/519386994, the step took 100s on Linux + Python3.6.
At https://travis-ci.org/HazyResearch/fonduer/jobs/519448788, it took 80s on Linux + Python 3.6.
20 seconds is not significant speedup but small incremental improvements are collectively important.

PR #246, opened 2019-04-12
Not upgrade PostgreSQL for OSX to speedup TravisCI

HiromuHota:


HiromuHota:
`if [ "$TRAVIS_OS_NAME" = "osx" ]; then ls /usr/local/opt/readline/lib/; fi` gives

libhistory.8.0.dylib	libhistory.dylib	libreadline.a
libhistory.8.dylib	libreadline.8.0.dylib	libreadline.dylib
libhistory.a		libreadline.8.dylib	pkgconfig

HiromuHota:
Readline was upgraded to 8.0.0 to satisfy Poppler's dependency on it.
Okay, may not worth spending more time on this.

HiromuHota:
The latest .travis.yml can reduce a job time on osx by 6-7 mins (32mins -> 25mins), which I think is significant.

PR #247, opened 2019-04-22
Create Entity class

HiromuHota:
I found it cumbersome to make learning-related util functions, e.g., `entity_confusion_matrix`, `entity_level_f1`, for each and every application.
To make these functions as a first-class citizen in the package, I think the "Entity" has to be standardized.

I think these are principles and requirements for `Entity` class:
- 1-N relationship between `Entity` and `Mention` (i.e., one entity -  N mentions).
- An instance of `Entity` can be associated with multiple types of `Mention`, e.g., `SpanMention`, `CellMention`, `FigureMention`.
- An instance of `Entity` has to be uniquely identifiable.
- The way an instance of `Entity` is identified should be customizable.


HiromuHota:
Close for now for inactivity.

PR #248, opened 2019-04-25
Incorporate entity_confusion_matrix as a first-class citizen

HiromuHota:
This PR will become ready for review once I submit a corresponding PR for tutorials.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -12,6 +12,7 @@ Added
 
 Changed
 ^^^^^^^
+* `@HiromuHota`_: Incorporate entity_confusion_matrix as a first-class citizen and rename it to confusion_matrix because it can be used both entity-level and mention-level.
 * `@HiromuHota`_: Separate Spacy#_split_sentences_by_char_limit to test itself.
 * `@HiromuHota`_: Refactor the custom sentence_boundary_detector for readability and efficiency.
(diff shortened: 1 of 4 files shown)


HiromuHota:
Snorkel-metal has an API for a confusion matrix with different input/output.
https://github.com/HazyResearch/metal/blob/c24e3772e25ac6d0917b8b7af4c1bcb92928f84a/metal/analysis.py#L217-L242

To avoid confusion, Fonduer should use the Metal's API if Metal has one.
@SenWu Thoughts?

HiromuHota:
Snokel also has a slightly different API.
https://github.com/HazyResearch/snorkel/blob/1d16ae6a6dba0f51c60d5a4179576d2d89134160/snorkel/learning/classifier.py#L93-L122

SenWu:
> Snorkel-metal has an API for a confusion matrix with different input/output.
> https://github.com/HazyResearch/metal/blob/c24e3772e25ac6d0917b8b7af4c1bcb92928f84a/metal/analysis.py#L217-L242
> 
> To avoid confusion, Fonduer should use the Metal's API if Metal has one.
> @SenWu Thoughts?

I would suggest to have Fonduer own implementation because of the different needs from Fonduer user.

HiromuHota:
Thanks for you suggestion!
I believe Snorkel's implementation is more similar to Fonduer's because Snorkel's returns a tuple of TP, FP, TN, FN.

PR #253, opened 2019-05-03
Remove useless data_model_utils.tabular.same_document

HiromuHota:


SenWu:
`make docs` failed. Can you remove the `same_document` related document as well?

HiromuHota:
I removed leftover imports. `make docs` ran w/o error locally.

PR #255, opened 2019-05-06
Visualizer test

HiromuHota:
I reverted CHANGELOG b/c this PR is not to change a previous change. This should be amended by a different PR.
I reverted visualizer.py and respect that the argument pdf_path is a directory where documents are stored.


requirements-dev.txt
```diff
@@ -1,5 +1,6 @@
 black>=18.9b0
 flake8
+ipython
```
SenWu: do we need this for visualization test?

HiromuHota: Yes, because of the following import: https://github.com/HazyResearch/fonduer/blob/bef4217e2a40a37bd8d0f9b7803a4e0d467eadbe/src/fonduer/utils/visualizer.py#L8

PR #257, opened 2019-05-07
Move ipython into setup.py b/c it is not a dev tool

HiromuHota:


setup.py
```diff
@@ -30,6 +30,7 @@
         "tqdm>=4.26.0, <5.0.0",
         "treedlib>=0.1.1, <0.2.0",
         "wand>=0.4.4, <0.5.0",
+        "ipython",
```
SenWu: can we have the version requirement as well?

HiromuHota: At this moment, I have no idea about the boundary between those work and those don't.
But, I don't think it is a good idea to pose a strict version constraint on ipython like > 7.5.0 (latest).
Maybe `ipython>=7.0.0` is a good compromise?

HiromuHota: Hold on. `ipython>=7.0.0` might give a wrong message to users that Fonduer should work with ipython of 7.0.0 or greater, which might not be the case.
I'd go with without a version constraint until we find the correct one.

SenWu: Make sense!

PR #268, opened 2019-05-17
Assign a section for mention spaces

HiromuHota:
MentionSpaces are conceptually different from Mention/CandidateExtractor.
So it would be a good idea to assign a whole section for MentionSpaces like Matchers.

### DIFF
--- a/docs/user/candidates.rst
+++ b/docs/user/candidates.rst
@@ -19,9 +19,53 @@ Core Objects
 
 These are Fonduer_'s core objects used for Mention and Candidate extraction.
 
-.. automodule:: fonduer.candidates
+.. autoclass:: fonduer.candidates.MentionExtractor
     :members:
     :inherited-members:
+    :show-inheritance:
+
+.. autoclass:: fonduer.candidates.CandidateExtractor
+    :members:
+    :inherited-members:
+    :show-inheritance:
+
+MentionSpaces
+-------------
+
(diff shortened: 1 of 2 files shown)


docs/user/candidates.rst
```diff
@@ -19,11 +19,55 @@ Core Objects
...
 
+.. autoclass:: fonduer.candidates.CandidateExtractor
+    :members:
+    :inherited-members:
+    :show-inheritance:
+
+MentionSpaces
+-------------
+
+A *MentionSpace* defines a space where mentions lie on. You can use a pre-defined child
```
lukehsiao: > A *MentionSpace* defines a space where mentions lie on.

I wonder if we can clarify this statement. Maybe something like, "defines the space of mentions considered by Fonduer"? Any thoughts here?

HiromuHota: I think the first statement should tell the concept of MentionSpace without talking about any particular implementation. Any following ones can talk about Fonduer's implementation.
How about this?
"A *MentionSpace* defines the space of mentions, i.e., the set of all possible mentions."

lukehsiao: That sounds great! If you update that I think this PR is ready to merge.

HiromuHota: Thanks! I modified the 1st statement, (and the 2nd a little bit).

PR #271, opened 2019-05-24
Clarify the effect of longest_match_only when used with Union, Intersect, or Inverse

HiromuHota:
I realized that `longest_match_only` of a child matcher is ignored when used with `Union`.
If you want to disable `longest_match_only`, you have to add `longest_match_only=False` to `Union` instead of an individual matcher.

This is true for other matcher operators like `Inverse` and `Intersect` too, but I'm not sure how this affects the result.
@SenWu, @lukehsiao, any thoughts?

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -70,6 +70,7 @@ Changed
 * `@HiromuHota`_: Make load_lang_model private as a model is internally loaded during init.
 * `@HiromuHota`_: Add a unit test for ``Parser`` with tabular=False.
   (`#261 <https://github.com/HazyResearch/fonduer/pull/261>`_)
+* `@HiromuHota`_: Now ``longest_match_only`` of ``Union``, ``Intersect``, and ``Inverse`` override that of child matchers.
 
 Removed
 ^^^^^^^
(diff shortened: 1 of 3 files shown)


lukehsiao:
Even if we don't change the underlying behavior, it would definitely be good to clarify what the behavior is in the documentation like you've started to do here.

HiromuHota:
A question: why does neither `Intersect` nor `Inverse` inherit `_NgramMatcher` while `Union` inherits it?

SenWu:
Imo they should not, they matchers should be more general to support different types of mentions

HiromuHota:
I agree with the idea of "they matchers should be more general to support different types of mentions".
However, I also think that they should be consistent, meaning `Union` should not inherit `_NgramMatcher` when `Intersect` and `Inverse` does not.

HiromuHota:
So, `Union` now inherits `_Matcher`, making it consistent with `Intersect` and `Inverse`.

This change (along with the change to the "seen_spans(mentions)" logic) made the effect of `longest_match_only` consistent among these.
Before:
- `longest_match_only` of `Union` overrides that of child matchers.
- `longest_match_only=True` of `Intersect` and `Inverse` used to have no effect. When `Intersect` and `Inverse` are used, the matches are not always longest (because _is_subspan always False).

Now:
- `longest_match_only` of `Union`, `Intersect` and `Inverse` override that of child matchers.

PR #272, opened 2019-05-24
Remove the default value "<NB>" from delim of SimpleTokenizer

HiromuHota:
Removing the default value "\<NB\>", which used to be used for Stanford CoreNLP.
Because Fonduer now mainly uses spaCy instead of Stanford CoreNLP, "\<NB\>" should not be the default.
You can still specify delim="\<NB\>" if they need it.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -99,6 +99,8 @@ Changed
   (`#306 <https://github.com/HazyResearch/fonduer/issues/306>`_)
 * `@HiromuHota`_: Pin PyTorch on 1.1.0 to align with Snorkel of 0.9.X.
 * `@HiromuHota`_: Depend on psycopg2 instead of psycopg2-binary as the latter is not recommended for production.
+* `@HiromuHota`_: Change the default value for ``delim`` of ``SimpleParser`` from "<NB>" to ".".
+  (`#272 <https://github.com/HazyResearch/fonduer/pull/272>`_)
 
 Removed
 ^^^^^^^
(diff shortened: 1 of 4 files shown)


SenWu:
@HiromuHota let me when you think this can be merged into master?

HiromuHota:
This cannot be merged now as the test fails.

The test fails because we are not passing anything to `delim`, meaning `delim` becomes `None`.
According to https://docs.python.org/3/library/stdtypes.html#str.split,
> If sep is not specified or is None, a different splitting algorithm is applied: runs of consecutive whitespace are regarded as a single separator, and the result will contain no empty strings at the start or end if the string has leading or trailing whitespace.

This behaviour may not be what we want for the SimpleTokenizer.

I think this is not just a bug fix, but more like a design question if `delim` is not `<NB>`.
How should SimpleTokenizer split text?

SenWu:
@HiromuHota, good point!

The design goal for `SimpleTokenizer` is to parse content and output (yield) parsed all sentences, where the `delim` is to split the content in to sentences and then for each sentence it uses `.strip()` to do tokenization.

Sen

HiromuHota:
So the question is "what would be the default delimiter to split text into sentences?"
I'd say that is "." (period), and changed the patch accordingly.

I'd admit that this sentencizer is too simple. For example, "U.S.A." will be split into three sentences.
@SenWu any thought?

SenWu:
@HiromuHota This is a good point. SimpleTokenizer is designed to help parse simple data input without NLP attributes. On usage is plain text where each line is a sentence and the goal is to extract information for such sentences. So in the SimpleTokenizer, I think Fonduer will take a delimiter as input to split the data into sentences.

HiromuHota:
So... is the current patch good to go or needs some change?

SenWu:
Yes, good to merge.

PR #274, opened 2019-06-04
Make Visualizer#display_candidates mention-type argnostic

HiromuHota:
This PR assumes a mention has attributes like `top`, `left`, `top`, `right`, and `bottom`.
This PR stays as a draft until for example every mention has those attributes.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -129,6 +129,7 @@ Fixed
 * `@HiromuHota`_: Fix the issue that Labeler.apply with docs instead of split fails.
   (`#340 <https://github.com/HazyResearch/fonduer/pull/340>`_)
 * `@HiromuHota`_: Make mention/candidate_subclasses and their objects picklable.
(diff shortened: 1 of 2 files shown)


HiromuHota:
I was gonna let mentions of any type have top, left, right, bottom attributes, but I'm not working on this direction now.
If and only if an app code adds those attributes to extracted mentions, Visualizer#display_candidates works.
Merge this if this PR is helpful.

SenWu:
Can we fix the test and merge this PR?

HiromuHota:
Resolved the conflict. Please merge if it looks good.

lukehsiao:
Can merge this after conflict is resolved.

HiromuHota:
Let me see what was wrong during OSX test.
There is a chance that the same root cause as https://github.com/HazyResearch/fonduer/pull/347 is affecting here.

HiromuHota:
https://travis-ci.org/HazyResearch/fonduer/jobs/604026106 was not able to reproduce the error #351 is trying to avoid.
Let me just clean this PR and please merge it once tests are finished.

PR #279, opened 2019-06-26
Add Labeler#upsert_keys

HiromuHota:
Like I added `Featurizer#upsert_keys` (#258), I'd like to add `Labeler#upsert_keys`.
The use case is to deploy a generative model (instead of a discriminative model) somewhere different from where you train this generative model.


SenWu:
This is great idea. Can we have a test for this as well?

HiromuHota:
Yes, I'll add another commit for a test.

PR #280, opened 2019-07-01
WIP: Add page argument to get_pdf_dim in case pages have different dimensions

HiromuHota:
Some pdfs have pages in different dimensions like `BC546A_Series_B14-521026.pdf`.
This patch accommodates such a case with API backward compatibility.


SenWu:
Should this test only on lInux env?

HiromuHota:
Tests failed on mac because MagickWand is not installed.

```
src/fonduer/utils/visualizer.py:9: in <module>
    from wand.color import Color
../../../miniconda/envs/test-environment/lib/python3.6/site-packages/wand/color.py:9: in <module>
    from .api import MagickPixelPacket, library
../../../miniconda/envs/test-environment/lib/python3.6/site-packages/wand/api.py:210: in <module>
    'Try to install:\n  ' + msg)
E   ImportError: MagickWand shared library not found.
E   You probably had not installed ImageMagick library.
E   Try to install:
E     brew install freetype imagemagick
```

Do you want to install freetype and imagemagick on mac for test?

SenWu:
If we can have that, that will be AWESOME!

PR #285, opened 2019-07-23
docs: update changelog to follow keepachangelog standard

lukehsiao:
Specifically, we want to make sure that each section detailed the types
of changes are presented in the same order. The order we use is:

- Added
- Changed
- Deprecated
- Removed
- Fixed
- Security

See https://keepachangelog.com/en/1.0.0/.


SenWu:
Can merge when tests pass.

HiromuHota:
+1

Issue #260, opened 2019-05-14
Separate VisualLinker from Parser


HiromuHota:
**Is your feature request related to a problem? Please describe.**

If you use HTML files as the primary source, you would print them as PDF files in a pre-preprocessing.
You would use `HTMLDocPreprocessor`, and `Parser` with `visual=True` and `pdf_path` args, which activates the visual linker that links a HTML file and the corresponding PDF file to get the visual information.

On the other hand, when you start from CSV, TSV, or Text files, you would not use `VisualLinker` (ie `Parser` with `visual=False` and `pdf_path=None` args).

If you start from PDFs, you would use `pdftotext` to extract the text layer or apply OCR to them when they have no text layer. In either case, you would get HTML files with textual and visual information. However, the current implementation of Fonduer cannot use the visual information embedded in HTML files.

**Describe the solution you'd like**

1. Separate VisualLinker from `Parser` and make it as a helper tool, and remove `pdf_path` arg from `Parser`
2. Change the meaning of `visual=True` from activating the visual linker to extracting the visual information embedded in HTML files (or any type of primary source files).

The helper tool takes a HTML file and the corresponding PDF file, runs the visual linker, embeds the visual information to the input HTML file, and returns it.

**Describe alternatives you've considered**

The visuallinker could be migrated to `HTMLDocPreprocessor`, but I'd suggest to make it a complete separate helper tool to make the Fonduer pipeline (starting from docpreprocessor) simpler.

**Additional context**

The current meaning of `visual=True` is different from other modals like `lingual=True`: the former means activating the visual linker, while the latter means extracting something embedded in the source files, which is confusing.

Once the solution I described above is implemented, `visual=True`, `lingual=True`, `tabular=True` all mean extracting the corresponding modal information from the (primary) source files.

HiromuHota:
Resolved by #287.

PR #287, opened 2019-07-24
Pluggable visual linker

HiromuHota:
This patch addresses #260 by adding `vizlink` as an argument to `Parser` to be able to plug a custom visual linker.
Unless otherwise specified, `VisualLinker` will be used by default for backward-compatibility.

### DIFF
--- a/src/fonduer/parser/parser.py
+++ b/src/fonduer/parser/parser.py
@@ -229,7 +229,7 @@ class ParserUDF(UDF):
             if self.visual:
                 # Use the provided pdf_path if present
                 self.pdf_path = pdf_path if pdf_path else self.pdf_path
-                if not self.vizlink.valid_pdf(document.name):
+                if not self.vizlink.is_linkable(document.name):
                     warnings.warn(
                         (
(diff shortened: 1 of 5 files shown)


src/fonduer/parser/parser.py
```diff
@@ -48,6 +47,9 @@ class Parser(UDFRunner):
     :param tabular: Whether to include tabular information in the parse.
     :param visual: Whether to include visual information in the parse.
         Requires PDFs for each input document.
+    :param vizlink: A custom visual linker that inherits
+        :class:`VisualLinker <fonduer.parser.visual_linker.VisualLinker>`.
+        Unless otherwise specified, :class:`VisualLinker` will be used. Default None.
```
lukehsiao:
```suggestion
        Unless otherwise specified, :class:`VisualLinker` will be used.
```

This was a little confusing to say that something will be used, but defaults to None. I'd just say the above.

HiromuHota:
Thanks for reviewing and your suggestion.

Issue #290, opened 2019-07-30
Type hints


HiromuHota:
**Is your feature request related to a problem? Please describe.**

I'm always frustrated when I have to look at the source codes to check the type of arguments/return.

**Describe the solution you'd like**

1. Type hints (PEP484) are written to source codes like

```python
def greeting(name: str) -> str:
    return 'Hello ' + name
```

2. (Eventually) enforce type checking during pre-commit

For example by [flake8-mypy](https://pypi.org/project/flake8-mypy/)

**Describe alternatives you've considered**

Depending on the editor (PyCharm, etc.), type/rtype documentation like below gives you type hinting.
However, I'm not sure this is equivalent to the type hints (PEP484).

```
def greeting(name):
    """
    greeting

    :param name: description
    :type name: type description
    :return: description
    :rtype: type description
    """
    return 'Hello ' + name
```

**Additional context**

SenWu:
This is definitely a good idea!!! Let's do this!

SenWu:
Any suggestions?

HiromuHota:
[PyAnnotate](https://github.com/dropbox/pyannotate) looks good. This is a tool that auto-generates PEP-484 annotations.
This has been developed by the same team, at Dropbox, who has originally developed mypy.
Let me try PyAnnotate for a while.

By the way, I don't know exactly what you mean by the following, but I don't think we need a folder 'type' for type hinting. Let me know if I'm mistaken.
> One more question about typing: do you think it's a good idea to have type folder to include all types in Fonduer?

HiromuHota:
https://github.com/Instagram/MonkeyType could be another option.

HiromuHota:
So basically, both MonkeyType and PyAnnotate do the same thing: 1. collect types at runtime and 2. insert collected types to the source code.
I ran `pytest tests` with each but none of the two traced the calls of Fonduer methods because `pytest tests` ran against the Fonduer package installed in site-packages, which is ignored by MonkeyType and PyAnnotate by default.
Fortunately, MonkeyType takes care of such a case by specifying the package to be traced in an environment variable (see [here](https://monkeytype.readthedocs.io/en/latest/faq.html#i-run-tests-against-my-package-installed-in-site-packages-and-i-get-no-traces)).

The following works with Fonduer,

1. Collect types
```bash
$ pip install pytest-monkeytype
$ export MONKEYTYPE_TRACE_MODULES=fonduer
$ cd fonduer # move to the root directory
[...]

PR #291, opened 2019-07-31
Pluggable lingual parser

HiromuHota:
This PR adds 
-  `LingualParser` class, which any lingual parser like `Spacy` should inherit from
- `lingual_parser` as an argument to `Parser` to be able to plug a custom lingual parser
Unless otherwise specified, `Spacy` will be used by default for backward-compatibility.


src/fonduer/parser/lingual_parser.py
```diff
@@ -0,0 +1,46 @@
...
+
+        :param text: text to be split
+        :type text: str
+        :return: A generator of dict that is used as `**kwargs` to instantiate
+            :class:`Sentence <fonduer.parser.models.Sentence>`.
+        """
+        raise NotImplementedError()
+
+    def enrich_sentences_with_NLP(
+        self, all_sentences: Iterable[Sentence]
```
SenWu: Just use `sentences` ?

src/fonduer/parser/spacy_parser.py
```diff
@@ -15,7 +17,7 @@
     raise Exception("spaCy not installed. Use `pip install spacy`.")
 
 
-class Spacy(object):
+class Spacy(LingualParser):
```
SenWu: Rename to `SpacyParser`?

src/fonduer/parser/lingual_parser.py
HiromuHota: Renamed into `sentences` by 0d480b893a3758447579c40c3ae6b252957a31dc.

src/fonduer/parser/spacy_parser.py
HiromuHota: Renamed as suggested at b07538a706aa03a33ebda4510d2144309da7dc28.

HiromuHota:
Thanks for the suggestions. I'm currently adopting all of them.
Will push them when I'm done.

HiromuHota:
Rebased and (force) pushed. Let me know if there is anything.

CHANGELOG.rst
```diff
@@ -57,6 +57,9 @@ Added
         from fonduer.parser import Parser
         parser = Parser(session, vizlink=CustomVisualLinker())
 
+* `@HiromuHota`_: Add `LingualParser`, which any lingual parser like `Spacy` should inherit from,
+  and add `lingual_parser` as an argument to `Parser` to be able to plug a custom lingual parser.
+
```
SenWu: Can we update CHANGELOG as well?

HiromuHota: I added some changes to CHANGELOG but I didn't include changes that do not affect how to use Fonduer like additional tests, doc section, etc.

HiromuHota:
I believe the following question is for #290. Let me know if I'm wrong.

> One more question about typing: do you think it's a good idea to have type folder to include all types in Fonduer?

SenWu:
Yes, you are right!

PR #307, opened 2019-09-17
Use the official name "beautifulsoup4" instead of an alias "bs4"

### LINKED ISSUE #306: Inappropriate dependency on bs4 (instead of beautifulsoup4)
**Describe the bug**

This is not a bug, but bs4 is a dummy package and not supposed to be used on purpose.

https://pypi.org/project/bs4/ says,
> This package ensures that if you type pip install bs4 by mistake you will end up with Beautiful Soup.

**Expected behavior**
Dependency on beautifulsoup4 with an appropriate range of versions.
Currently any version of beautifulsoup4 satisfies the dependency requirement.

**Error Logs/Screenshots**

```
 $ pipdeptree -p fonduer
fonduer==0.7.1
  - bs4 [required: >=0.0.1,<0.1.0, installed: 0.0.1]
    - beautifulsoup4 [required: Any, installed: 4.7.1]
      - soupsieve [required: >=1.2, installed: 1.9.1]
...
```

HiromuHota:
This will resolve #306.

The latest version of beautifulsoup4 is now 4.8.0.
Though I couldn't find any breaking changes at 4.8.0 as far as [CHANGELOG](https://bazaar.launchpad.net/~leonardr/beautifulsoup/bs4/view/head:/CHANGELOG) is concerned, let's stay 4.7.1 for now.
If you are confident to include 4.8.0 and higher, please let me know.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -71,6 +71,8 @@ Changed
 * `@HiromuHota`_: Add a unit test for ``Parser`` with tabular=False.
   (`#261 <https://github.com/HazyResearch/fonduer/pull/261>`_)
 * `@HiromuHota`_: Now ``longest_match_only`` of ``Union``, ``Intersect``, and ``Inverse`` override that of child matchers.
+* `@HiromuHota`_: Use the official name "beautifulsoup4" instead of an alias "bs4".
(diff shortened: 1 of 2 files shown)


HiromuHota:
Thanks for your prompt action.

PR #313, opened 2019-09-23
Type annotations-5 (related to #290)

HiromuHota:
Another series of commits for type annotations (#290).
This reduces the number of mypy complains from 151 to 105.
This PR remains draft until #312 gets merged.


HiromuHota:
The test failed because https://download.pytorch.org/whl/torch_stable.html does not list the 1.2.0 version currently.
I believe this is temporarily happening to resolve https://github.com/pytorch/pytorch/issues/26510.

PR #320, opened 2019-09-28
Use Snorkel v0.9.X instead of snorkel-metal

### LINKED ISSUE #310: Make -1 as abstain to be consistent with snorkel v0.9.0
**Describe the bug**

Currently, Fonduer uses 0 for ABSTAIN as Snorkel of 0.7.0 (and Metal) does so.
However, Snorkel as of v0.9.0 uses -1 for ABSTAIN.
To leverage the recent advances at Snorkel v0.9.0, I think Fonduer should also use -1 for ABSTAIN to get prepared to future upgrade to Snorkel v0.9.0 (or higher).

**Expected behavior**

Fonduer uses -1 for ABSTAIN to be consistent with snorkel v0.9.0.
After that, Fonduer uses snorkel of v0.9.0 (or higher) instead of metal.

**Additional context**

The Snorkel team's discussion about ABSTAIN standardization: https://github.com/HazyResearch/metal/issues/65

HiromuHota:
To leverage the new features of Snorkel, I'd like Fonduer to keep up with it.
This will close #310.


tests/e2e/test_e2e.py
```diff
@@ -421,14 +421,14 @@ def test_e2e(caplog):
     ]
     labeler.update(split=0, lfs=[stg_temp_lfs_2, ce_v_max_lfs], parallelism=PARALLEL)
     assert session.query(Label).count() == 6478
-    assert session.query(LabelKey).count() == 16
+    assert session.query(LabelKey).count() == 19
```
SenWu: Why there are 3 more LabelKey here?

tests/e2e/test_e2e.py
```diff
@@ -364,10 +364,10 @@ def test_e2e(caplog):
     L_train_gold = labeler.get_gold_labels(train_cands, annotator="gold")
     assert L_train_gold[0].shape == (3493, 1)
 
-    gen_model = LabelModel(k=2)
-    gen_model.train_model(L_train[0], n_epochs=500, print_every=100)
+    gen_model = LabelModel()
+    gen_model.fit(L_train=L_train[0].toarray(), n_epochs=500, log_freq=100)
 
-    train_marginals = gen_model.predict_proba(L_train[0])
+    train_marginals = gen_model.predict_proba(L_train[0].toarray())
```
SenWu: Why use dense format here? It will cause potential overhead?

tests/shared/hardware_lfs.py
```diff
@@ -8,9 +8,9 @@
     overlap,
 )
 
-ABSTAIN = 0
-FALSE = 1
-TRUE = 2
+ABSTAIN = -1
```
SenWu: How about moving these constants to hardware_utils and import from there?

src/fonduer/utils/utils_udf.py
```diff
@@ -29,6 +29,8 @@
 # integer to ensure that it won't conflict with a user's split value.
 ALL_SPLITS = "ALL"
 
+ABSTAIN = -1
```
SenWu: Should this be in the labeler since it's labeler specific?

tests/e2e/test_e2e.py
HiromuHota: The total number of unique LFs, `len(set(stg_temp_lfs + ce_v_max_lfs + stg_temp_lfs_2))`, is 20, while `session.query(LabelKey).all()` does not include `LF_temp_outside_table`, as a result `session.query(LabelKey).count()` is 19.

Like `LF_operating_row` returns `0` (ABSTAIN) for all candidates when `ABSTAIN==0`, `LF_temp_outside_table` returns `0` (FALSE) for all of them when `ABSTAIN==-1 (FALSE==0)`.

Because `Labeler` skips records with `value==0`, no labels for `LF_operating_row` is recorded.

tests/e2e/test_e2e.py
HiromuHota: This is due to the API change at snorkel. They are now mainly use a dense format for matrices.

tests/shared/hardware_lfs.py
HiromuHota: Thanks for your suggestion. Done at 47d94ed0d377d0caf83c87b9ae901532431cb71a.

src/fonduer/utils/utils_udf.py
HiromuHota: I thought so at first, but soon realized that it'd cause a circular imports between labeler and utils_udf.

tests/e2e/test_e2e.py
HiromuHota: The Snorkel team standardized their coding strategy and API as follows:
> label matrix is dense outside of compute modules
> Internal computation for LF analysis and label model is with shifted sparse matrix: csr_matrix(L + 1)

https://github.com/snorkel-team/snorkel/pull/1309#issue-300125859

src/fonduer/utils/utils_udf.py
HiromuHota: Done at 6e7f16965e298dab1fe2702ef968cbe66b842333.

tests/e2e/test_e2e.py
HiromuHota: To avoid the conversion (sparse to dense) overhead, `labeler.get_label_matrices` should return a dense format as it's been standardized at Snorkel like below:
> label matrix is dense outside of compute modules

SenWu: There is no reason to return dense matrix here since the label matrix supposes to be sparse.

tests/e2e/test_e2e.py
SenWu: Using -1 as ABSTAIN is not a computation efficient way since most of the values are ABSTAIN.

SenWu: Good catch!

SenWu:
Besides the several design choices of snorkel, LGTM, can merge. @lukehsiao, thoughts?

lukehsiao:
It does seem strange that snorkel switched to non-zero ABSTAIN and dense matrices. Can we chat with them about that? I would imagine that that results in some performance regressions? 

If it doesn't seem like a major issue, I'm fine with the change.

HiromuHota:
@lukehsiao I also wonder the reason behind the switch. I'd appreciate if you could ask them about it.

tests/e2e/test_e2e.py
SenWu: One more thought here, the total number of LabelKey should be 20 since the 0 means negative (no longer means abstain) and labeler should not ignore it.

HiromuHota: Good catch!
To avoid any confusion and computational inefficiency, how about stick with snorkel's policy?
> ABSTAIN is now represented by None (LFs), -1 (label matrix), or 0 (LFAnalysis, except for when we use L and then it's -1).
https://github.com/snorkel-team/snorkel/pull/1309#pullrequestreview-265185691

So -1 for user-facing labels and 0 for internal computation and database?

setup.py
```diff
@@ -20,7 +20,7 @@
         "numpy>=1.11, <2.0",
         "pyyaml>=4.2b1, <5.0",
         "scipy>=1.1.0, <2.0.0",
-        "snorkel-metal>=0.4.1, <0.5.0",
+        "snorkel>=0.9.0, <0.10.0",
```
SenWu: This is not the latest version of snorkel. The latest version fixed the networkx version.

setup.py
```diff
@@ -20,7 +20,7 @@
...
         "networkx>=2.2, <2.4",
```
SenWu: remove this.

src/fonduer/learning/classifier.py
```diff
@@ -403,7 +403,7 @@ def predict(self, X, b=0.5, pos_label=1, return_probs=False):
         self.logger.info(f"Using positive label class {pos_label} with threshold {b}")
 
         Y_pred = np.array(
-            [pos_label if p[pos_label - 1] > b else 3 - pos_label for p in Y_prob]
+            [pos_label if p[pos_label] > b else pos_label - 1 for p in Y_prob]
```
SenWu: The new label starts from 0. So, in binary case the other case is `1 - pos_label`?

src/fonduer/learning/classifier.py
```diff
@@ -452,7 +452,7 @@ def score(
         if self.cardinality == 2:
             # Either remap or filter out unlabeled (0-valued) test labels
             if set_unlabeled_as_neg:
-                Y_test[Y_test == 0] = 3 - pos_label
+                Y_test[Y_test == 0] = pos_label - 1
```
SenWu: `1 - pos_label`

src/fonduer/supervision/labeler.py
```diff
@@ -37,6 +37,11 @@
 
 logger = logging.getLogger(__name__)
 
+# Snorkel standardized ABSTAIN as follows:
+# ABSTAIN is None (LFs), -1 (label matrix), or 0 (LFAnalysis)
```
SenWu: This sentence is not clear.

src/fonduer/supervision/labeler.py
```diff
@@ -356,30 +361,38 @@ def _after_apply(
...
         :param cand_lists: The candidates to get labels for.
         :type cand_lists: List of list of candidates.
-        :return: A list of MxN sparse matrix where M are the candidates and N is the
+        :return: A list of MxN dense matrix where M are the candidates and N is the
             labeling functions.
-        :rtype: list[csr_matrix]
+        :rtype: list[np.ndarray]
         """
-        return get_sparse_matrix(self.session, LabelKey, cand_lists)
+        return [
```
SenWu: Can we make a wrapper function in utils for this?

src/fonduer/supervision/labeler.py
```diff
@@ -356,30 +361,38 @@ def _after_apply(
...
-        :rtype: list[csr_matrix]
+        :rtype: list[np.ndarray]
         """
-        return get_sparse_matrix(self.session, GoldLabelKey, cand_lists, key=annotator)
+        return [
+            m.toarray() - 1  # convert to {-1, 0, ..., k-1}
+            for m in get_sparse_matrix(
+                self.session, GoldLabelKey, cand_lists, key=annotator
+            )
+        ]
```
SenWu: Can we make a wrapper function in utils for this?

src/fonduer/supervision/labeler.py
```diff
@@ -356,30 +361,38 @@ def _after_apply(
...
 
-    def get_label_matrices(self, cand_lists: List[List[Candidate]]) -> List[csr_matrix]:
-        """Load sparse matrix of Labels for each candidate_class.
+    def get_label_matrices(self, cand_lists: List[List[Candidate]]) -> List[np.ndarray]:
```
SenWu: Can we add an option to support returning sparse/dense matrix here?

src/fonduer/supervision/labeler.py
```diff
@@ -356,30 +361,38 @@ def _after_apply(
 
     def get_gold_labels(
         self, cand_lists: List[List[Candidate]], annotator: Optional[str] = None
```
SenWu: Can we add an option to support returning sparse/dense matrix here?

setup.py
HiromuHota: Resolved by b85300a10fbc2e21220d68953c245f0d67d08f90.

setup.py
HiromuHota: Resolved by b85300a10fbc2e21220d68953c245f0d67d08f90.

src/fonduer/learning/classifier.py
HiromuHota: Good catch! Resolved by 3ba6cf0048490e15161f3ae232735b812a7667eb.

src/fonduer/learning/classifier.py
HiromuHota: Resolved by 3ba6cf0048490e15161f3ae232735b812a7667eb.

src/fonduer/supervision/labeler.py
HiromuHota: How about this?
# Snorkel changed the label conventions: ABSTAIN is now represented by -1 (used to be 0)
# Details can be found at https://github.com/snorkel-team/snorkel/pull/1309

SenWu: Looks good! Also, add 0 index for classification task labels.

src/fonduer/supervision/labeler.py
HiromuHota: Which one do you mean?
- sparse (ABSTAIN as -1)/dense (ABSTAIN as -1)
- sparse (ABSTAIN as 0)/dense (ABSTAIN as -1)

src/fonduer/supervision/labeler.py
HiromuHota: See the change at 6428ddf8ea5df85a4168455173036f705e3be2c5.

src/fonduer/supervision/labeler.py
SenWu: The second one. We can have a wrapper and a unwrapper function for this.

HiromuHota: Please help me find a use case for such an option.
Once Fonduer shifts ABSTAIN to -1, users are forced to shift their labels too.
To avoid a confusion, I think anything that assumes ABSTAIN as 0 should not be exposed to users.

SenWu: Agree! I think it’s more clear to have a wrapper function (convert user behavior (ABSTAIN = -1) to fonduer internal behavior (ABSTAIN = 0)) and unwrapper function to do this then just hard code it. And I think this wrapper should not expose to user.

HiromuHota: Thanks. Will do.

src/fonduer/supervision/labeler.py
HiromuHota: Implemented at 7fc08d985a65e08ff117d4bc66d372fbd30546e9.

src/fonduer/supervision/labeler.py
HiromuHota: Implemented at 7fc08d985a65e08ff117d4bc66d372fbd30546e9.

src/fonduer/supervision/labeler.py
HiromuHota: Resolved by 7fc08d985a65e08ff117d4bc66d372fbd30546e9.

src/fonduer/supervision/labeler.py
HiromuHota: No adding such an option here.
Label matrix shifter and unshifter have been implemented at 7fc08d985a65e08ff117d4bc66d372fbd30546e9.

SenWu:
Can merge when conflict resolved.

Issue #311, opened 2019-09-20
So many "e2e/test_e2e.py::test_e2e" log messages during pytest


HiromuHota:
**Describe the bug**

This is not necessarily a bug, but so many "e2e/test_e2e.py::test_e2e" log messages started to appear during pytest.

**To Reproduce**

This starts at 1aa28bbd7b1d0956a4c18e85eaad6a57138ba437 (use torch of 1.2.0 for Travis and Docker).

**Expected behavior**

Only informative log messages should appear.

**Error Logs/Screenshots**
If applicable, add error logs or screenshots to help explain your problem.

Look at https://travis-ci.org/HazyResearch/fonduer/jobs/587176800

**Environment (please complete the following information):**
 - PyTorch: 1.2.0

HiromuHota:
These log messages appear right after 
```
  /home/travis/miniconda/envs/test-environment/lib/python3.6/site-packages/networkx/algorithms/components/connected.py:86: DeprecationWarning: connected_component_subgraphs is deprecated and will be removedin 2.2. Use (G.subgraph(c).copy() for c in connected_components(G))
    _warnings.warn(msg, DeprecationWarning)
```

Thus basically these are deprecation warnings.

HiromuHota:
This issue might be related: https://github.com/pytorch/pytorch/issues/24918

PR #329, opened 2019-10-11
Switch learning engine to Emmental

SenWu:


setup.py
```diff
@@ -16,15 +16,15 @@
     install_requires=[
         "beautifulsoup4==4.7.1",
         "editdistance>=0.5.2, <0.6.0",
+        "emmental @ git+https://github.com/SenWu/emmental.git@master",
```
lukehsiao: Ideally, I think we do want a stable versioned released before merging in.

SenWu: Agree! I will release a new version of Emmental soon.

setup.py
```diff
@@ -16,15 +16,15 @@
...
         "lxml>=4.2.5, <5.0.0",
         "numpy>=1.11, <2.0",
         "pyyaml>=4.2b1, <5.0",
         "scipy>=1.1.0, <2.0.0",
         "snorkel-metal>=0.4.1, <0.5.0",
         "spacy>=2.1.3, <2.2.0",
         "sqlalchemy[postgresql_psycopg2binary]>=1.2.12, <2.0.0",
-        "tensorboardX>=1.6, <2.0",
-        "torch>=1.1.0,<1.2.0",
+        "torch>=1.2.0,<1.3.0",
```
HiromuHota: Let travis install torch of 1.2.0 if you set "torch>=1.2.0,<1.3.0" here.

HiromuHota:
Please try to annotate types when your memory is fresh.

HiromuHota:
Use `pip install torch==1.3.0+cpu -f https://download.pytorch.org/whl/torch_stable.html` in travis to install a much lighter version of torch (travis does not provide GPU, does it?).
Details can be found at #309.

SenWu:
torch==1.3.0+cpu doesn’t exist last night?

HiromuHota:
Maybe but now 1.3.0+cpu is listed on https://download.pytorch.org/whl/torch_stable.html.
Are you sure about >=1.3.0? It's just released yesterday. We don't know its stability yet.

SenWu:
Great! It works now. I will use 1.3.0+cpu for linux. 1.3.0+cpu is the latest stable version. :)

.travis.yml
```diff
@@ -78,17 +77,12 @@ before_install:
...
-    if [ "${PYTHON}" = "3.7" ]; then
-      pip install https://download.pytorch.org/whl/cpu/torch-1.1.0-cp37-cp37m-linux_x86_64.whl
-    fi
+    pip install -q torch==1.3.0+cpu -f https://download.pytorch.org/whl/torch_stable.html
   fi
 
 install:
 # Install fonduer dev deps and check code style
-- make -s dev_extra
+- make -s dev_extra > /dev/null
```
HiromuHota: Please do not silent this line.
Without knowing the version of each dependent package, debugging would become difficult.
For example, it might be harder to find networkx 2.4 was installed here https://github.com/HazyResearch/fonduer/pull/335#issuecomment-543381163

SenWu: Agree! This is hacky to reduce the log size. Any suggestion there?

HiromuHota: I'd suggest to just use `make dev_extra` here (`-s` option does not seem to have any effect here).
If you worry about the log size, #311 produces way more logs because snorkel-metal is not compatible (more precisely, snorkel-metal is using deprecated APIs of torch) with torch>=1.2. That's why #319 pins torch on 1.1.0.

lukehsiao:
You probably want to update the Acknowledgement block in the README in this PR also.

SenWu:
Good catch! Will update this PR tonight.

CHANGELOG.rst
```diff
@@ -101,6 +101,90 @@ Added
 
 Changed
 ^^^^^^^
+* `@senwu`_: Switch to Emmental as the default learning engine.
+
+.. note::
+
+    Instead of owning Fonduer learning engine, we switch to Emmental which is a deep learning
+    framework for multi-task learning. This change can give Fonduer more power to support more
+    applications. Example usage:
```
lukehsiao: Suggest to reword to something like:

> Rather than maintaining a separate learning engine, we switch to Emmental, a deep learning framework for multi-task learning. Switching to a more general learning framework allows Fonduer to support more applications and multi-task learning. Example usage:

CHANGELOG.rst
```diff
@@ -101,6 +101,90 @@ Added
...
+    .. code:: python
+
+        import emmental
+
+        # Collect word counter from candidates
+        word_counter = collect_word_counter(train_cands)
+
+        emmental.init(fonduer.Meta.log_path)
+
+        # Geneate special tokens
```
lukehsiao: I personally think it would be nice to give more detailed comments here unless you think a user will understand this. Why are these special tokens needed? Why is it in the format in line 125? 

```suggestion
        # Generate special tokens
```

CHANGELOG.rst
```diff
@@ -101,6 +101,90 @@ Added
...
+        # Generate word embedding module
+        emb_layer = EmbeddingModule(
+            word_counter=word_counter, word_dim=300, specials=specials
+        )
+
+        # Only use the samples that have labels
+        diffs = train_marginals.max(axis=1) - train_marginals.min(axis=1)
+        train_idxs = np.where(diffs > 1e-6)[0]
+
+        # Build training dataloader
```
lukehsiao: Similarly, would be nice to have a brief mention of the why for these comments. I see that we're declaring a EmmentalDataLoader, but why do I need it? What does it to? This will help a user understand what this is replacing in the old code.

E.g., Create a dataloader to <purpose>

docs/user/config.rst
```diff
@@ -45,20 +45,7 @@ The default ``.fonduer-config.yaml`` configuration file is shown below::
         attention: True
         dropout: 0.1
         bidirectional: True
-        host_device: "CPU"
-        max_sentence_length: 100
       LogisticRegression:
         bias: False
```
lukehsiao: Can we add docs or a link to how to configure Emmental?

CHANGELOG.rst
```diff
@@ -3,6 +3,117 @@ Unreleased_
...
+
+        # Initialize Emmental. For customize Emmental, please check here: 
+        # https://emmental.readthedocs.io/en/latest/user/config.html
+        emmental.init(fonduer.Meta.log_path)
+
+        #######################################################################
+        # 1. Create task for each relations and EmmentalModel to learn those tasks.
+        #######################################################################
+
+        # Geneate special tokens which are used for LSTM model to locate mentions.
```
lukehsiao:
```suggestion
        # Generate special tokens which are used for LSTM model to locate mentions.
```

src/fonduer/learning/task.py
```diff
@@ -0,0 +1,192 @@
...
+    return sce_loss(intermediate_output_dict[module_name][0][active], label[active])
+
+
+def output(module_name: str, intermediate_output_dict: Dict[str, Any]) -> Tensor:
+    return F.softmax(intermediate_output_dict[module_name][0])
+
+
+def create_task(
+    task_names: Union[str, List[str]],
+    n_arites: Union[int, List[int]],
```
lukehsiao:
```suggestion
    n_arities: Union[int, List[int]],
```

src/fonduer/learning/task.py
```diff
@@ -0,0 +1,192 @@
...
+    mode: str = "mtl",
+    model: str = "LSTM",
+) -> List[EmmentalTask]:
+    """Create task from relation(s).
+
+    :param task_names: Relation name(s), If str, only one relation; If List[str],
+        multiple relations.
+    :type task_names: str, List[str]
+    :param n_arites: Relation arity of each relation.
+    :type n_arites: int, List[int]
```
lukehsiao:
```suggestion
    :type n_arities: int, List[int]
```

src/fonduer/learning/task.py
```diff
@@ -0,0 +1,192 @@
...
+    emb_layer: Optional[EmbeddingModule],
+    mode: str = "mtl",
+    model: str = "LSTM",
+) -> List[EmmentalTask]:
+    """Create task from relation(s).
+
+    :param task_names: Relation name(s), If str, only one relation; If List[str],
+        multiple relations.
+    :type task_names: str, List[str]
+    :param n_arites: Relation arity of each relation.
```
lukehsiao:
```suggestion
    :param n_arities: The arity of each relation.
```

src/fonduer/learning/task.py
```diff
@@ -0,0 +1,192 @@
...
+        raise ValueError(
+            f"Unrecognized model {model}. Only support {['LSTM', 'LogisticRegression']}"
+        )
+
+    config = get_config()["learning"][model]
+    logger.info(f"{model} model config: {config}")
+
+    if not isinstance(task_names, list):
+        task_names = [task_names]
+        n_arites = [n_arites]
```
lukehsiao: Switch all these to `n_arities`

src/fonduer/learning/task.py
```diff
@@ -0,0 +1,192 @@
...
+    :type n_features: int
+    :param n_classes: Number of classes for each task. (Only support classification
+        task now).
+    :type n_classes: int, List[int]
+    :param emb_layer: The embedding layer for LSTM. No need for LogisticRegression
+        model.
+    :type emb_layer: EmbeddingModule
+    :param mode: Model mode, defaults to "mtl".
+    :type mode: str
+    :param model: Model name (avaiable model: "LSTM", "LogisticRegression"),
```
lukehsiao:
```suggestion
    :param model: Model name (available models: "LSTM", "LogisticRegression"),
```

src/fonduer/learning/task.py
```diff
@@ -0,0 +1,192 @@
...
+    :type n_arities: int, List[int]
+    :param n_features: The multimodal feature set size.
+    :type n_features: int
+    :param n_classes: Number of classes for each task. (Only support classification
+        task now).
+    :type n_classes: int, List[int]
+    :param emb_layer: The embedding layer for LSTM. No need for LogisticRegression
+        model.
+    :type emb_layer: EmbeddingModule
+    :param mode: Model mode, defaults to "mtl".
```
lukehsiao: What are the options here? Just "mtl" and "stl"? Can we make it explicit so users don't have to guess?

lukehsiao: Actually, it seems this parameter isn't even used?

src/fonduer/learning/utils.py
```diff
@@ -107,3 +94,81 @@ def confusion_matrix(pred: Set, gold: Set) -> Tuple[Set, Set, Set]:
...
+
+
+def mention_to_tokens(
+    mention: Mention, token_type: str = "words", lowercase: bool = False
+) -> List[str]:
+    """
+    Extract tokens from the mention
+
+    :param mention: mention object.
+    :param token_type: token type that wants to extract.
```
lukehsiao: What token types are there?

lukehsiao:
I asked ReadTheDocs about the memory issue with PyTorch. This was their response:

> Thanks for reporting this to us.
>
> This is a known issue. It seems that there is some memory leaking with that package and pip because it consumes a lot of memory to be installed. Unfortunately, there is not too much we can do at this time.
>
> I'm taking the suggestion to include pytorch as a system package. Although, we don't upgrade the Docker images too often, so it pytorch will probably be outdated immediately.
>
> I'm not sure how are you using pytorch to build the docs, but in case that you want to document the code itself, you may want to give a try to https://sphinx-autoapi.readthedocs.io/en/latest/ that documents the code without executing it, so there is no need to install a lot of dependencies.

That said, I doubt the autoapi extension really doesn't need the package to be present, as it sounds essentially the same as the autodoc extension we are using. We should still look into it.

lukehsiao:
Looking further, it seems the autoapi extension was built by readthedocs directly to help address the limitations of autodocs:

> Sphinx AutoAPI provides "autodoc" style documentation for multiple programming languages without needing to load, run, or import the project being documented.
>
>In contrast to the traditional Sphinx autodoc, which is Python-only and uses code imports, AutoAPI finds and generates documentation by parsing source code.

https://github.com/readthedocs/sphinx-autoapi

So, migrating over to this may save us the hassle of dealing with imports in the future.

lukehsiao:
`autoapi` lacks the ability to show inherited members. See `autoapi_options` in 

https://sphinx-autoapi.readthedocs.io/en/latest/reference/config.html

This is an autodoc feature we are heavily using, so it doesn't seem like a good drop-in replacement.

PR #335, opened 2019-10-17
Fix deadlock error during Labeler.apply and Featurizer.apply

HiromuHota:
This PR fix #328.

The root cause of #328 is that multiple processes (when PARALLEL >= 2) try to update the key tables (LabelKey and FeatureKey) at the same time with almost the same content.
This patch creates a method `UDFRunner.after_apply`, which will be executed by a single process.

PR #341, opened 2019-10-21
Make after_apply method as private

HiromuHota:
I should have made it private at #335.


HiromuHota:
`UDFRunner.after_apply` was left unchanged. My bad!
I change it and force pushed.

PR #345, opened 2019-10-22
Travis w/o conda and only on Python 3.7

HiromuHota:
Motivated by frequent failures during conda installation, this PR removes the conda dependency.

Conda was helpful to install different versions of Python, 3.6 and 3.7.
Without conda, only one version of Python (ie 3.7) will be tested on.

It is worth noting that Ubuntu is upgraded to Bionic because poppler-utils for Xenial is 0.41.0, which is not recommended.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,3 @@
-language: generic
-dist: xenial
 cache:
   directories:
   - $HOME/.cache/pip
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -33,19 +26,28 @@ notifications:
...
     sudo sed -i 's/max_connections = 255/max_connections = 100/g' /etc/postgresql/10/main/postgresql.conf
     sudo service postgresql restart
     # Install ImageMagick library for Wand
     sudo apt-get install libmagickwand-dev ghostscript
     sudo rm -rf /etc/ImageMagick-6/policy.xml
+    # Install poppler-utils
+    sudo apt install -q -y poppler-utils
   else
     OS=MacOSX-x86_64
+    pip install virtualenv
```
SenWu: pip3?

HiromuHota: Thanks. pip actually works fine on OSX.
For better consistency, pip3 is better but Makefile also needs fixing.
Do you mind if I fix Makefile too?

SenWu: How about we use alias to use `pip` for all the time?

HiromuHota: Okay. Will do.

Issue #10, opened 2018-02-10
Handle psycopg2 renaming


lukehsiao:
>....venv/lib/python3.5/site-packages/psycopg2/__init__.py:144: UserWarning: The psycopg2 wheel package will be renamed from release 2.8; in order to keep installing from binary please use "pip install psycopg2-binary" instead. For details see: <http://initd.org/psycopg/docs/install.html#binary-install-from-pypi>.

We need to see if we just switch to the binary or if there are other better options.

PR #349, opened 2019-10-25
Depend on psycopg2 instead of psycopg2-binary

HiromuHota:
As stated in https://pypi.org/project/psycopg2/ that
> The binary package is a practical choice for development and testing but in production it is advised to use the package built from sources.

also in http://initd.org/psycopg/docs/install.html#binary-packages that
> If you are the maintainer of a publish package depending on psycopg2 you shouldn’t use ‘psycopg2-binary’ as a module dependency. For production use you are advised to use the source distribution.

Fonduer should depend on psycopg2 and build it from source.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -66,6 +66,8 @@ before_install:
     pg_ctl -D /usr/local/var/postgres start; sleep 5
     createuser -s postgres
     # poppler-utils is already installed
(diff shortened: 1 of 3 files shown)


HiromuHota:
This is a rework of the fix for #10.

PR #347, opened 2019-10-23
Merge _apply_mt and _apply_st for maintainability

HiromuHota:
This PR becomes ready for review when a dependent PR #342 is merged.

PR #352, opened 2019-10-28
Merge _apply_mt and _apply_st for maintainability

HiromuHota:
This PR merges `_apply_mt` and `_apply_st` for better maintainability.

Even when parallelism==1, this PR will spawn another process.
This doesn't sound computationally efficient, but should be fine because parallelism==1 means that the task itself is small.

This is a rework of #347.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,0 +1,2 @@
+[run]
+concurrency = multiprocessing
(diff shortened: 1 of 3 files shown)


HiromuHota:
I still have to figure out the reason for
> Coverage decreased (-13.6%) to 68.964%

HiromuHota:
Now I found it.
> Code executed by worker pools or child processes is not tracked.

https://stackoverflow.com/questions/28297497/python-code-coverage-and-multiprocessing

PR #367, opened 2019-12-02
Mypy on fonduer.learning.* too

HiromuHota:
This is related to #290.

Now that Fonduer has Emmental, it is time to check typing at fonduer.learning.*.

### DIFF
--- a/src/fonduer/learning/dataset.py
+++ b/src/fonduer/learning/dataset.py
@@ -99,13 +99,14 @@ class FonduerDataset(EmmentalDataset):
                 ]
                 s = mark_sentence(mention_to_tokens(candidate[i]), args)
                 self.X_dict[f"m{i}"].append(
-                    torch.LongTensor(
+                    torch.tensor(
                         [
                             self.word2id[w]
                             if w in self.word2id
                             else self.word2id["<unk>"]
                             for w in s
-                        ]
+                        ],
+                        dtype=torch.long,
                     )
                 )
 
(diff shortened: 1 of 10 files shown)


HiromuHota:
Travis failed because the current code is not conformal to Mypy 0.750, which was released on Nov. 29.

src/fonduer/learning/dataset.py
```diff
@@ -99,29 +99,31 @@ def _map_to_id(self) -> None:
...
                         self.features.indptr[i] : self.features.indptr[i + 1]
-                    ]
+                    ],
+                    dtype=torch.long,
                 )
                 + 1
             )
             self.X_dict["feature_weight"].append(
-                torch.Tensor(
+                torch.tensor(
```
HiromuHota: This change broke the test.

HiromuHota: The error is RuntimeError: expected backend CPU and dtype Double but got backend CPU and dtype Float and it happened at `return (w.unsqueeze(2) * self.weight(x)).sum(dim=1)` (https://github.com/SenWu/emmental/blob/v0.0.4/src/emmental/modules/sparse_linear_module.py#L66).

HiromuHota: `self.weight` is `nn.Embedding` and `self.weight(x)` is a tensor of dtype==float32.

HiromuHota: When an error occurred, `w.unsqueeze(2)` was a tensor of dtype==double.

HiromuHota: Multiplying two tensor of different dtype (float and double) caused the RuntimeError.

HiromuHota: torch.Tensor is an alias for the default tensor type (torch.FloatTensor); thus, float32.
torch.tensor() constructor infers dtype from data. Seems like dtype==double was inferred.
Should have explicitly specified dtype.

PR #368, opened 2019-12-04
No database dependency during inference/serving

### LINKED ISSUE #316: No database dependency during inference/serving

I think a Fonduer-based app lifecycle has three phases: development, training, and serving.
During development, you may have many iterations of re-definition of mention/candidate subclasses, labeling functions, which involve re-parsing, re-extraction, re-labeling, etc.
I understand that it is important to persist intermediate results to a database to save time.
Depending on the size of training dataset, you may need to persist intermediate results during training too.

On the other hand, there is no need to persist intermediate results to a database during inference/serving.
The dependency on database (postgres) makes a Fonduer-based app less portable and less scalable.

I'd like no database dependency during inference/serving.

sqlite is easier to install than postgres, but still there is no need to persist intermediate results to a database.

**Additional context**

Related to #137.

HiromuHota:
This will close #316 once it's finished.

### DIFF
--- a/src/fonduer/candidates/candidates.py
+++ b/src/fonduer/candidates/candidates.py
@@ -18,6 +18,7 @@ from sqlalchemy.orm import Session
 from fonduer.candidates.models import Candidate, Mention
 from fonduer.parser.models.document import Document
 from fonduer.utils.udf import UDF, UDFRunner
+from fonduer.utils.utils import get_set_of_stable_ids
 
 logger = logging.getLogger(__name__)
 
@@ -280,17 +281,9 @@ class CandidateExtractorUDF(UDF):
                     for mention in candidate_class.mentions
                 ]
             )
-            # Construct a set of stable ids of candidates.
-            set_of_stable_ids = set()
-            if hasattr(doc, candidate_class.__tablename__ + "s"):
-                set_of_stable_ids.update(
-                    set(
-                        [
-                            tuple(m.context.get_stable_id() for m in c)
-                            for c in getattr(doc, candidate_class.__tablename__ + "s")
-                        ]
-                    )
-                )
+            # Get a set of stable_ids of candidates.
+            set_of_stable_ids = get_set_of_stable_ids(doc, candidate_class)
+
             for cand in cands:
 
                 # Apply throttler if one was given.
--- a/src/fonduer/candidates/candidates.py
+++ b/src/fonduer/candidates/candidates.py
@@ -15,7 +15,6 @@ from typing import (
 )
 
 from sqlalchemy.orm import Session
-from sqlalchemy.sql import select
 
 from fonduer.candidates.models import Candidate, Mention
 from fonduer.parser.models.document import Document
@@ -259,12 +258,11 @@ class CandidateExtractorUDF(UDF):
         super().__init__(**kwargs)
 
     def apply(  # type: ignore
-        self, doc: Document, clear: bool, split: int, **kwargs: Any
+        self, doc: Document, split: int, **kwargs: Any
     ) -> Iterator[Candidate]:
         """Extract candidates from the given Context.
 
         :param doc: A document to process.
-        :param clear: Whether or not to clear the existing database entries.
         :param split: Which split to use.
         """
         logger.debug(f"Document: {doc}")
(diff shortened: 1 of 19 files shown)


HiromuHota:
Although it needs some code cleanup, it's now functioning.
As can be seen in https://github.com/HiromuHota/fonduer-mlflow/tree/fix/316, a database is not needed during inference.
I'll do some code cleaning and make it ready for review shortly.

HiromuHota:
The number of commits was reduced to 18, but still a lot to review.
I'll be splitting them into smaller PRs so that it is easier to review.

src/fonduer/parser/parser.py
```diff
@@ -249,13 +249,13 @@ def __init__(
 
     def apply(  # type: ignore
         self, document: Document, pdf_path: Optional[str] = None, **kwargs: Any
-    ) -> Iterator[Sentence]:
+    ) -> Document:
         # The document is the Document model
         text = document.text
 
         # Only return sentences, if no exceptions occur during parsing
```
lukehsiao: Let's update this comment, too. Since we are returning a document now, not sentences.

lukehsiao:
I know this is unrelated to this specific PR, but I wonder if there is a way to eliminate the ~1k warnings we get in test_e2e on travis.

src/fonduer/parser/parser.py
HiromuHota: Good catch. I update it by simply removing the comment and adding a docstring.

HiromuHota:
If I remember correctly, the ~1k warnings come from Snorkel where it uses deprecated APIs of PyTorch.

src/fonduer/candidates/candidates.py
```diff
@@ -283,6 +280,17 @@ def apply(  # type: ignore
                     for mention in candidate_class.mentions
                 ]
             )
+            # Construct a set of stable ids of candidates.
+            set_of_stable_ids = set()
+            if hasattr(doc, candidate_class.__tablename__ + "s"):
```
SenWu: Can we have a documentation here to mention why `+ "s"`?

src/fonduer/candidates/candidates.py
```diff
@@ -283,6 +280,17 @@ def apply(  # type: ignore
...
+                set_of_stable_ids.update(
+                    set(
+                        [
+                            tuple(m.context.get_stable_id() for m in c)
+                            for c in getattr(doc, candidate_class.__tablename__ + "s")
```
SenWu: Same here.

src/fonduer/candidates/candidates.py
```diff
@@ -314,16 +322,18 @@ def apply(  # type: ignore
...
-                if not clear:
-                    q = select([candidate_class.id])
-                    for key, value in list(candidate_args.items()):
-                        q = q.where(getattr(candidate_class, key) == value)
-                    candidate_id = self.session.execute(q).first()
-                    if candidate_id is not None:
-                        continue
+                    candidate_args[arg_name] = cand[j][1]
+
+                stable_ids = tuple(
```
SenWu: Can we have a global func to construct stable id? It's a little bit messy to have stable id constructor everywhere.

src/fonduer/candidates/mentions.py
```diff
@@ -577,46 +563,61 @@ def __init__(
...
-    def apply(  # type: ignore
-        self, doc: Document, clear: bool, **kwargs: Any
-    ) -> Iterator[Mention]:
+    def apply(self, doc: Document, **kwargs: Any) -> Document:
         """Extract mentions from the given Document.
 
         :param doc: A document to process.
-        :param clear: Whether or not to clear the existing database entries.
         """
+        # Construct a dict of stable ids of contexts.
```
SenWu: Is it possible to move this to the utils?

src/fonduer/candidates/candidates.py
HiromuHota: I moved these stable_id constructors to `fonduer.utils.utils`.

src/fonduer/candidates/candidates.py
HiromuHota: I added a comment to explain the reason.
See https://github.com/HazyResearch/fonduer/pull/368/files#diff-dd77b82c8178c68913da115c82628c3dR56.

src/fonduer/candidates/candidates.py
HiromuHota: See https://github.com/HazyResearch/fonduer/pull/368/files#diff-dd77b82c8178c68913da115c82628c3dR56.

PR #378, opened 2020-01-24
Get `cands_list` and `records` without db access

HiromuHota:
This is the 3rd part of the PR #368.


HiromuHota:
Travis failed due to an flake8 error.
Fixed it and (force) pushed.

PR #379, opened 2020-02-11
Use the term "label model" instead of "generative model"

HiromuHota:
Snorkel as of v0.9.3 uses "Label Model" and "End Model" instead of "Generative Model" and "Discriminative Model".

This is because, which I'm not really sure, the label model is not a generative model any more as of v0.9.3 (AAAI-19 paper).
Correct me if I'm wrong.

### DIFF
--- a/tests/e2e/test_e2e.py
+++ b/tests/e2e/test_e2e.py
@@ -374,10 +374,10 @@ def test_e2e():
     L_train_gold = labeler.get_gold_labels(train_cands, annotator="gold")
     assert L_train_gold[0].shape == (3493, 1)
 
-    gen_model = LabelModel()
-    gen_model.fit(L_train=L_train[0], n_epochs=500, log_freq=100)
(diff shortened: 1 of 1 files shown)


HiromuHota:
I was wrong about what I said
> the label model is not a generative model any more as of v0.9.3 (AAAI-19 paper).

For the record, @ajratner said in a different thread (https://spectrum.chat/snorkel/help/is-label-model-still-a-generative-model~b36cb359-ce97-4edf-9590-b64b6a35c1d9),
> We would still call this a generative model. First, with the AAA'19 formulation, we can still directly compute P(\lambda, Y) (since we can directly observe P(\lambda)). Second, and more broadly: we call it a generative model because we still assume a (simple) model of the data generation process.
Regarding the terminology shift: we just changed this because we thought it would be a bit clearer and more direct re: what the models are actually intended to do vs. how they do it.

PR #381, opened 2020-03-29
Fix mention extraction to use mention classes, not data model classes

SenWu:
When building multimodal candidates, Fonduer should extract multimodal mention which should be XXMention class (e.g., ParagraphMention, DocumentMention) instead of data model class (e.g., Paragraph, Document). This PR fixes this issue.


CHANGELOG.rst
```diff
@@ -130,6 +130,10 @@ Changed
   (`#316 <https://github.com/HazyResearch/fonduer/issues/316>`_)
   (`#368 <https://github.com/HazyResearch/fonduer/pull/368>`_)
 
+Fixed
+^^^^^
+* `@senwu`_: Fix multimodal mention doesn't return mention class.
```
lukehsiao: Let's make this more descriptive. Is this specific to multimodal? Should it be something more general like "fix X so that Y returns mention classes, rather than data model classes"?

src/fonduer/parser/models/utils.py
```diff
@@ -15,13 +15,30 @@ def construct_stable_id(
...
-    doc_id, _, parent_doc_char_start, _ = split_stable_id(parent_context.stable_id)
+
+    doc_id, _, idx = split_stable_id(parent_context.stable_id)
+
+    # Caption, document, figure, paragraph, section, table
+    if len(idx) == 1:
+        parent_doc_start = idx[0]
+        return f"{doc_id}::{polymorphic_type}:{parent_doc_start}"
+    # Cell
+    elif len(idx) == 3:
```
lukehsiao: This feels like a magic number to me. Why is this a cell? Is there a way to make this code easier to read than relying on len(idx)?

CHANGELOG.rst
lukehsiao: After reading the code, perhaps we call this:

Fix mention extraction to operate on mention classes, not data model

src/fonduer/parser/models/utils.py
```diff
@@ -15,13 +15,35 @@ def construct_stable_id(
...
-    doc_id, _, parent_doc_char_start, _ = split_stable_id(parent_context.stable_id)
-    start = parent_doc_char_start + relative_char_offset_start
-    end = parent_doc_char_start + relative_char_offset_end
-    return f"{doc_id}::{polymorphic_type}:{start}:{end}"
 
-
-def split_stable_id(stable_id: str) -> Tuple[str, str, int, int]:
+    doc_id, type, idx = split_stable_id(parent_context.stable_id)
+
+    if polymorphic_type in [
```
lukehsiao: Awesome.

PR #382, opened 2020-03-30
Release v0.8.0

lukehsiao:
Bumping up the emmental dependency to v0.0.6 in an effort to clean up our dependencies for v0.8.0.


setup.py
```diff
@@ -16,15 +16,15 @@
...
         "beautifulsoup4==4.7.1",
         "editdistance>=0.5.2, <0.6.0",
-        "emmental==0.0.4",
+        "emmental>=0.0.5, <0.1.0",
         "lxml>=4.2.5, <5.0.0",
         "numpy>=1.11, <2.0",
         "pyyaml>=5.1, <6.0",
         "scipy>=1.1.0, <2.0.0",
-        "snorkel>=0.9.2, <0.10.0",
+        "snorkel @ git+https://github.com/lukehsiao/snorkel.git@luke-patch",
```
lukehsiao: This one is an attempt to resolve our "dependency hell". Snorkel may eventually update their dependencies, and we can go back to mainline snorkel once they do.

HiromuHota: Please do not release Fonduer until this line reverts back to the released version of Snorkel.
This is because `https://github.com/lukehsiao/snorkel.git@luke-patch` may not be available for a long-term.

lukehsiao: Yes, we will not :+1:

lukehsiao: I don't even want to merge this PR until this line is gone.

lukehsiao, line 24: Need to resolve this before merging. There are 3 options, as I see it.

1. Snorkel bumps up dependencies. I don't know how likely this is to happen (https://github.com/snorkel-team/snorkel/issues/1558).
2. We fork snorkel to HazyResearch org as a more "permanent" fork. Not ideal.
3. We downgrade our use of pytorch to be compatible. Also not ideal.

Any other suggestions?

lukehsiao:
It seems to me we can have a nice clean release of v0.8.0 if we get a new release of emmental with the adjusted dependencies, snorkel with adjusted dependencies, and then we can release fonduer v0.8.0. We can then release new version of things like our hardware application pinned to fonduer v0.8.0. Snorkel is the piece we don't have control over.

setup.py
```diff
@@ -16,15 +16,15 @@
     install_requires=[
         "beautifulsoup4==4.7.1",
         "editdistance>=0.5.2, <0.6.0",
-        "emmental==0.0.4",
+        "emmental>=0.0.5, <0.1.0",
```
lukehsiao, line 19: We should release emmental first, then bump this up to v0.0.6.

setup.py
HiromuHota: I'd suggest a combination of 1 and 3.
Downgrade pytorch 1.1.0 until Snorkel changes their dependency on pytorch to 1.3.1 or higher.

lukehsiao:
An additional reason to not fork: we cannot release on PyPI will pinned to a direct dependency, or we'll get an error like:

```
HTTPError: 400 Client Error: Invalid value for requires_dist. Error: Can't have direct dependency: 'snorkel @ git+https://github.com/lukehsiao/snorkel.git@luke-patch' for url: https://test.pypi.org/legacy/
```

lukehsiao:
https://github.com/snorkel-team/snorkel/pull/1566

We will be able to have snorkel pytorch compatibility with their v0.9.5 release :tada:

lukehsiao:
Snorkel v0.9.5 has been released. Now, we just need to release emmental v0.0.6.

lukehsiao:
Snorkel introduced tensorboard as a dependency with the <2.0.0 limit:

```
ERROR: snorkel 0.9.5 has requirement tensorboard<2.0.0,>=1.14.0, but you'll have tensorboard 2.2.0 which is incompatible.
```
emmental wants >= 2.0.1.  We will need to address this conflict as well.

lukehsiao:
We still see:
```
ERROR: snorkel 0.9.5 has requirement scikit-learn<0.22.0,>=0.20.2, but you'll have scikit-learn 0.22.2.post1 which is incompatible.
ERROR: snorkel 0.9.5 has requirement tensorboard<2.0.0,>=1.14.0, but you'll have tensorboard 2.2.0 which is incompatible.
```

Even though scikit-learn v0.21.0 and tensorboard v1.15.0 satisfy all of the requirements. This is confusing to me...

SenWu:
Make sure we remove all `Travis` keyword in the repo.

lukehsiao:
I believe all that is left now is the actual release process (updating CHANGELOG, versions, tagging, etc.).

PR #386, opened 2020-04-03
Use Python installed by default

HiromuHota:
On macOS, Python installed by default (3.7) is used.
actions/setup-python@v1 cannot be used on macos because "Can't connect to HTTPS URL because the SSL module is not available." error happens when trying to pip install.
Not confirmed, but this is probably because brew install openssl@1.1 as imagemagick dependency and breaks a link Python installed actions/setup-python@v1 and openssl.

### DIFF
--- a/.github/workflows/pythonpackage.yml
+++ b/.github/workflows/pythonpackage.yml
@@ -97,6 +97,10 @@ jobs:
     - uses: actions/checkout@v2
     - name: Install dependencies for macOS
       run: |
+        pip3 install virtualenv
+        virtualenv venv
+        source venv/bin/activate
+
         brew update
         brew install swig mecab mecab-unidic
         # Use unidic-mecab dictionary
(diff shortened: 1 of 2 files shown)


.github/workflows/pythonpackage.yml
```diff
@@ -95,17 +93,17 @@ jobs:
...
-    - name: Update Python ${{ matrix.python-version }} tools
-      run: python -m pip install --upgrade pip setuptools wheel
+    # actions/setup-python@v1 cannot be used on macos (see #386)
     - name: Install dependencies for macOS
       run: |
+        pip3 install virtualenv
+        virtualenv venv
+        VIRTUAL_ENV=`pwd`/venv
+        echo "::set-env name=VIRTUAL_ENV::$VIRTUAL_ENV"
+        echo "::add-path::$VIRTUAL_ENV/bin"
```
lukehsiao, line 105: Are these a workaround for a problem sourcing the virtualenv?

HiromuHota: Yes. Otherwise, you have to source the virtualenv every step like I did at f5b0b5a.
And you would need a virtualenv when you use the system python (the one installed by default), otherwise you will get a permission error, which you can avoid by adding `--user` option to every pip command.

lukehsiao:
Mind if I squash these into a single commit when I merge?

HiromuHota:
> Mind if I squash these into a single commit when I merge?

I don't mind at all. Please do.

PR #388, opened 2020-04-07
switch 0-index

SenWu:
Previously we use 1-index since it has better system performance in labeling with most of the labels are ABSTAIN. While 1-index is not user friendly and 0-index is commonly used in machine learning. So we propose to switch back to 0-index in both Emmental and Fonduer.


lukehsiao:
Before we merge, let's make a better commit message. I can take care of this via a squash/merge, or you can rebase and make a commit message. Something that includes why the switch is necessary would be good (perhaps with a link to the changelog of emmental).

PR #389, opened 2020-04-07
test the 0-index label in valid set and remove the log folder after t…

SenWu:
1. test the 0-index label in valid set.
2. remove the log folder after the test.

### DIFF
--- a/tests/e2e/test_e2e.py
+++ b/tests/e2e/test_e2e.py
@@ -392,7 +392,7 @@ def test_e2e():
 
     # Training config
     config = {
-        "meta_config": {"verbose": True},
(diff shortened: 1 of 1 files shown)


tests/e2e/test_e2e.py
```diff
@@ -76,9 +77,10 @@ def test_e2e():
     PARALLEL = 2
 
     max_docs = 12
+    dirpath = "temp_test_e2e"
```
lukehsiao, line 80: Looks like by default log_dir is tempfile.gettempdir(). Do, rather than specifying log_folder, you can just let it go to the temp directory and not worry about clean-up.

PR #392, opened 2020-04-10
Upload to coveralls.io only if 'push' on 'HazyResearch/fonduer'

### LINKED ISSUE #391: GitHub Actions fail on forked repository
**Describe the bug**

GitHub Actions fail on a forked repository because "Upload to coveralls.io" step fails.

**To Reproduce**
1. Fork HazyResearch/fonduer
2. Push any commit to your forked repository
3. GitHub Actions fails

**Expected behavior**

"Upload to coveralls.io" runs only on HazyResearch/fonduer

**Error Logs/Screenshots**

image

HiromuHota:
This will close #391 .

### DIFF
--- a/.github/workflows/pythonpackage.yml
+++ b/.github/workflows/pythonpackage.yml
@@ -78,6 +78,7 @@ jobs:
         # Use 127.0.0.1 instead of localhost (#351)
         POSTGRES_HOST: 127.0.0.1
     - name: Upload to coveralls.io
+      if: github.event_name == 'push' && github.repository == 'HazyResearch/fonduer'
       run: |
           coverage combine
           coveralls
@@ -166,6 +167,7 @@ jobs:
         # libiomp5-dylib-but-found-libiomp5-dylib-already-initial
         KMP_DUPLICATE_LIB_OK: true
     - name: Upload to coveralls.io
+      if: github.event_name == 'push' && github.repository == 'HazyResearch/fonduer'
       run: |
           coverage combine
           coveralls


HiromuHota:
Please see the step is skipped on a forked repository.
https://github.com/HiromuHota/fonduer/actions/runs/75405791.

HiromuHota:
This PR fails.
https://github.com/HazyResearch/fonduer/actions/runs/75436780

> coveralls.exception.CoverallsException: Not on TravisCI. You have to provide either repo_token in .coveralls.yml or set the COVERALLS_REPO_TOKEN env var.

- The "Upload to coveralls.io" step runs because this PR is against 'HazyResearch/fonduer'
- But, this PR fails because a token for coveralls is not available.

@lukehsiao any idea?

lukehsiao:
Not sure. The configuration sets the env var, and the secret is in the repo:

image

This might be due to the secret not being passed correctly?

https://help.github.com/en/actions/configuring-and-managing-workflows/creating-and-storing-encrypted-secrets

HiromuHota:
Yes, it is clearly mentioned in https://help.github.com/en/actions/configuring-and-managing-workflows/creating-and-storing-encrypted-secrets#using-encrypted-secrets-in-a-workflow that 
> With the exception of GITHUB_TOKEN, secrets are not passed to the runner when a workflow is triggered from a forked repository.

HiromuHota:
Or maybe "Coveralls GitHub Action" works better.
https://github.com/marketplace/actions/coveralls-github-action

HiromuHota:
"Coveralls GitHub Action" only supports LCOV file (`lcov.info`), while coverage.py, used by Fonduer, generates `.coverage`.
https://github.com/coverallsapp/github-action/issues/30

lukehsiao:
Hmmm, I'm fine with switching to codecov, there is no particular reason we are using coveralls.

SenWu:
I think codecov is easier solution here.

HiromuHota:
Let's fix this now. We can discuss if we should switch to codecov in a different issue/pr.

AndreMiras:
Before doing the switch give that one a try, I've just built a coveralls action that works for pull requests https://github.com/marketplace/actions/coveralls-python

PR #407, opened 2020-04-27
Add a custom MLflow model for Fonduer

### LINKED ISSUE #259: Serializable Fonduer model

I develop a Fonduer-based app locally on my laptop.
Once it's done, I'd like to package [the whole Fonduer pipeline](https://fonduer.readthedocs.io/en/latest/user/getting_started.html#the-fonduer-pipeline) (parsing, extraction, featurization, and classification) and deploy it to a remote place to serve.
However, a Fonduer-based app is not easy to package hence not easy to deploy.

Add a Fonduer model class that is 
1. Serializable (e.g., a class with `save` and `load` member methods like below)

```
class FonduerModel:
    def save(path_to_save):
    def load(path_to_load):
```

2. Capable of executing any phase of the Fonduer pipeline
3. (Hopefully) Manageable by MLflow

I can create one or more of python scripts that do all the phase, package them, and deploy it.
This is cumbersome because the python script has to include many things (matchers, mention_classes, mention_spaces, candidate_classes, etc.) and it is not obvious what should be included for serving.

**Additional context**

I'd like to make Fonduer more deployable and servable.
I've been testing MLflow to package a Fonduer-based app and found it was difficult to do so when there is no serializable Fonduer model.

HiromuHota:
This will close #259.


lukehsiao:
Do you want to include this in v0.8.2?

HiromuHota:
Although this PR would not introduce a breaking change, v0.8.2 does not have to wait for this.
I'm targeting at v0.9.0.

src/fonduer/utils/fonduer_model.py
```diff
@@ -0,0 +1,398 @@
+import logging
```
SenWu, line 1: This is good start.

One high-level comment here is: can restructure this giant file into a module of fonduer (e.g. packaging) and split this file into small files?

docs/user/packaging.rst
```diff
@@ -0,0 +1,15 @@
+Packaging
+=========
+
+You can package the whole Fonduer pipeline (parsing, extraction, featurization, and classification) and deploy it to a remote place to serve.
```
SenWu, line 4: It would be awesome to have more content about how to use it and when is a good time to use it? Just to share some idea there or maybe another tutorial for this?

CHANGELOG.rst
```diff
@@ -1,6 +1,13 @@
 Unreleased_
 -----------
 
+Added
+^^^^^
+* `@HiromuHota`_: Add a custom MLflow model that allows you to package a Fonduer model.
```
SenWu, line 6: Awesome!

src/fonduer/utils/fonduer_model.py
```diff
@@ -0,0 +1,398 @@
...
+    :param candidate_extractor: self-explanatory
+    :param conda_env: A dictionary representation of a Conda environment.
+    :param code_paths: A list of local filesystem paths to Python file dependencies,
+        or directories containing file dependencies. These files are prepended to the
+        system path when the model is loaded.
+    :param model_type: the model type, either "discriminative" or "generative",
+        defaults to "discriminative".
+    :param labeler: a labeler, defaults to None.
+    :param lfs: a list of list of labeling functions.
+    :param gen_models: a list of generative models, defaults to None.
```
SenWu, line 220: We use `label_model ` now?

tests/e2e/test_e2e.py
```diff
@@ -607,6 +602,35 @@ def test_e2e():
 
     assert f1 > 0.7
 
+    # Save a fonduer model
```
SenWu, line 605: It would be better to have separate tests for ml_flow.

tests/e2e/test_e2e.py
```diff
@@ -196,10 +198,6 @@ def test_e2e():
     temp_ngrams = MentionNgramsTemp(n_max=2)
     volt_ngrams = MentionNgramsVolt(n_max=1)
 
-    Part = mention_subclass("Part")
```
SenWu, line 199: This is awesome! We can also eliminate this part in incremental test as well.

src/fonduer/utils/fonduer_model.py
HiromuHota: I agree with moving this into `fonduer.packaging`.
I think I can move `F_matrix` and `L_matrix` into other files, but this would be the best thing I could do.
It seems to be the convention to have all the functions (e.g., `log_model`, `save_model`, wrapper, `_get_default_conda_env`) in a single file.
See https://github.com/mlflow/mlflow/blob/7312255f38a8c06181a9e3dfc0e9e08b45e21ea6/mlflow/pytorch/__init__.py

Let me see what I can do.

docs/user/packaging.rst
HiromuHota: Thanks you for your feedback.
I'll add more details to the docs, also I'll update one of the existing tutorials instead of adding another one.

src/fonduer/utils/fonduer_model.py
HiromuHota: Good catch!
With https://github.com/HazyResearch/fonduer/pull/379, Fonduer has switched to `label_model`.

HiromuHota: Can we also switch from "disc model" to "end model"?

CHANGELOG.rst
HiromuHota: Thanks.
Just to be clear, IMO I don't think Fonduer needs to be v0.9.X for this PR since this PR does not introduce any breaking changes.

tests/e2e/test_e2e.py
HiromuHota: Done by b9497adb606709537ed617c80cd37a89942fe6df.

src/fonduer/utils/fonduer_model.py
HiromuHota: Oh, now I can recall that I meant it.
I assume the following two scenarios.
1. a generative model as the end model
2. a discriminative model as the end model (a generative model is used purely as a label model)

In either case, the end model should be packaged, which is the generative model in the scenario 1 and the discriminative model in the scenario 2.

Snorkel changed the terminology (from gen_model vs. disc_model to label_model vs. end_model) to make them more clearer (what the models are actually intended to do vs. how they do it.)
From Snorkel perspective, this makes sense. However, from Fonduer (or an application that uses Snorkel) perspective, it may not if you use a generative model as the end model.

Having said, let's switch to label_model vs. end_model to avoid any confusion.

HiromuHota: label_model vs. emmental_model sounds clearer in Fonduer.

HiromuHota: Renamed `gen_model` and `disc_model`, to `label_model` and `emmental_model`, respectively.

tests/e2e/test_e2e.py
HiromuHota: Add additional unit tests.

docs/user/packaging.rst
HiromuHota: Added much more details in the doc.

Despite what I said earlier, I would not update an existing tutorial because it'd require separating some portion of a notebook into a module (like subclasses.py), which may distort what the original tutorial tries to convey.
Also creating another, new tutorial would be good, but requires much more effort.
I might create one in the future, but let me stop here.

src/fonduer/utils/fonduer_model.py
HiromuHota: I merged `F_matrix` (now `FonduerModel.convert_features_to_matrix`) with `fonduer.utils.utils_udf.get_sparse_matrix` to avoid duplicated codes.
fonduer_model.py still has 400+ lines of codes, but each function is not so long (the longest `save_model` has 70L without signature and docstring).

docs/user/packaging.rst
```diff
@@ -0,0 +1,93 @@
...
+        └── model.pkl  # the pickled Fonduer pipeline model.
+
+Currently, two types of classifiers are supported: :class:`EmmentalModel` (aka discriminative model) and :class:`LabelModel` (aka generative model).
+The following example shows how to package a Fonduer pipeline model that uses :class:`EmmentalModel` as a classifier.
+
+Example
+-------
+
+First, create a class that inherits :class:`FonduerModel` and implements :func:`_classify`.
+Then, put this class in a Python module like `my_fonduer_model.py` instead of in a Jupyter notebook or a long Python script as this module will be packaged.
```
SenWu, line 28: remove `long`

tests/shared/hardware_fonduer_model.py
```diff
@@ -0,0 +1,54 @@
+import pickle
```
SenWu, line 1: Should this file in the packaging folder or here? It seems like it's specific to the packaging test?

HiromuHota: This should be here because `HardwareFonduerModel._classify` uses it to unpickle "parts_by_doc_dict.pkl", which is specific to `HardwareFonduerModel`.

docs/user/packaging.rst
HiromuHota: I'll remove it.

HiromuHota: Resolved at 1fd13d11c61ac6b27a49bc72a8bcad9808725449.

docs/user/packaging.rst
```diff
@@ -0,0 +1,93 @@
...
+
+    .. code-block:: python
+        :caption: my_fonduer_model.py
+
+        class MyFonduerModel(FonduerModel):
+            def _classify(self, doc: Document) -> DataFrame:
+                # My implementation
```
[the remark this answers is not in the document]
lukehsiao: Maybe something like what is in `tests/shared/hardware_fonduer_model.py`?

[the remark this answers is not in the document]
HiromuHota: Thank you for your feedback!
At f03e6e65201044269c2e6148bb4d19c808d76bfa,
- added abstracted implementations at the docs to give users a bit more idea what they are supposed to implement.
- added references to two fully-functional examples.

PR #416, opened 2020-05-16
Improve documentation of preprocessors

### LINKED ISSUE #414: glob pattern is not documented as an acceptable path in DocPreprocessor
https://fonduer.readthedocs.io/en/latest/user/parser.html#preprocessors describes `path` to be filesystem path to file or directory to parse.
Actually, a glob pattern is acceptable path, but not documented.

### LINKED ISSUE #415: HTML files with the same base name in different sub-directories
**Describe the bug**

Two html files that are located as below are not parsed as expected.
```
html/folderA/doc.html
html/folderB/doc.html
```

You will get an error like below:
> DETAIL:  Key (stable_id)=(doc::document:0:0) already exists.

**To Reproduce**
1. Place two html files in path like aboe
2. Construct a preprocessor like  `preprocessor = HTMLDocPreprocessor("html/*/*")`

```python
>>> print(preprocessor.all_files)
['html/folderA/doc.html', 'html/folderB/doc.html']
```

3. Execute `parser.apply(preprocessor)`

**Expected behavior**

The two html files can be parsed.

**Error Logs/Screenshots**
If applicable, add error logs or screenshots to help explain your problem.

 - OS: Docker (hiromuhota/fonduer-tutorials:0.8.2)
 - PostgreSQL Version: 12.1
 - Poppler Utils Version: N/A
 - Fonduer Version: 0.8.2

**Additional context**

HiromuHota:
Describe a glob pattern as an acceptable form of a path, and describes the unique basename requirement.

`DocPreprocessor` has 5 subclasses and 2 of them (`HTMLDocPreprocessor` and `TextDocPreprocessor`)
inherit the __init__ method. To reduce the maintainance cost, autoclass_content = "both"
has been enabled so that these two subclasses's `__init__` docstring also inherits from their parent.
see below for more details of autoclass_content
https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#confval-autoclass_content

Fix #414.
Close #415.

### DIFF
--- a/src/fonduer/parser/preprocessors/doc_preprocessor.py
+++ b/src/fonduer/parser/preprocessors/doc_preprocessor.py
@@ -7,22 +7,22 @@ from fonduer.parser.models.document import Document
 
 
 class DocPreprocessor(object):
-    """
-    A generator which processes a file or directory of files into a set of
-    Document objects.
-
-    :param path: filesystem path to file or directory to parse.
-    :type path: str
-    :param encoding: file encoding to use (e.g. "utf-8").
-    :type encoding: str
-    :param max_docs: the maximum number of ``Documents`` to produce.
-    :type max_docs: int
-    :rtype: A generator of ``Documents``.
-    """
+    """An abstract class of a ``Document`` generator."""
 
     def __init__(
         self, path: str, encoding: str = "utf-8", max_docs: int = sys.maxsize
     ) -> None:
+        """
+        :param path: a path to file or directory, or a glob pattern. The basename
+            (as returned by ``os.path.basename``) should be unique among all files.
+        :type path: str
+        :param encoding: file encoding to use, defaults to "utf-8".
+        :type encoding: str, optional
+        :param max_docs: the maximum number of ``Documents`` to produce,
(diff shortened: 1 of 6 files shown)


src/fonduer/parser/preprocessors/doc_preprocessor.py
```diff
@@ -7,22 +7,25 @@
...
+    """An abstract class of a ``Document`` generator.
 
-    :param path: filesystem path to file or directory to parse.
-    :type path: str
-    :param encoding: file encoding to use (e.g. "utf-8").
-    :type encoding: str
-    :param max_docs: the maximum number of ``Documents`` to produce.
-    :type max_docs: int
-    :rtype: A generator of ``Documents``.
+    Unless otherwise stated by a subclass, it's assumed that one ``Document`` per file.
```
lukehsiao, line 12:
```suggestion
    Unless otherwise stated by a subclass, it's assumed that there is one ``Document`` per file.
```

lukehsiao:
Once the checks are fixed and tests pass we can merge.

PR #420, opened 2020-05-19
Fix minimum range difference for cols/rows in case of negative numbers

### LINKED ISSUE #413: get_neighbor_cell_ngrams with "absolute" parameter
It is hard to access a heading of a sub-table in a HTML where all sub-tables are in one `<table>` tag via a labeling function. `get_col_ngrams` can return the correct ngrams but without directions or distance. `get_neighbor_cell_ngrams` can return the direction, but does not consider columns which span multiple such as a heading to be neighboring.

I checked the code and adding an `absolute` flag to the `get_neighbor_cell_ngrams` function which is then passed to the function calls for row/col difference
```
row_diff = min_row_diff(sentence, root_cell, absolute=absolute)
col_diff = min_col_diff(sentence, root_cell, absolute=absolute)
```
could solve that problem. Instead of the `False` boolean default as it is currently.

Alternatively, I am quite often in need to access the table information of an ngram (e.g. which row/col is the sentence). The functions could provide an option to return the sentences instead of the ngrams. As far as I know I have not seen a way to get from the ngram back to the sentence(?).

**Additional context**
The example HTML in case needed for reproduction: 
[1_DJ Index 4-1 Peak.zip](https://github.com/HazyResearch/fonduer/files/4610678/1_DJ.Index.4-1.Peak.zip)

Example with an image: Access the heading "MID COLUMBIA" from the cell with the value `$235.33`. This is needed to know in which table the value is, as all this "sub-tables" are in one HTML table.
Screenshot from 2020-05-11 17-57-24

kaikun213:
Currently a difference of -2 column/row steps will override a difference of 1 or 0.
This leads to wrong behavior e.g. in for fields that span multiple columns/rows.

Closes #413

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -5,6 +5,11 @@ Changed
 ^^^^^^^
 * `@HiromuHota`_: Enabled "Type hints (PEP 484) support for the Sphinx autodoc extension."
 
+Fixed
+^^^^^
+* `@kaikun213`_: Fix bug in table range difference calculations.
+  (`#420 <https://github.com/HazyResearch/fonduer/pull/420>`_)
+
 0.8.2_ - 2020-04-28
 -------------------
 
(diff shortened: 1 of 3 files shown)


lukehsiao (changes requested):
Can you add a test case that demonstrates what this fixes? You can create a new file in `tests/utils/test_utils_table.py`. 

**Update**: I went ahead an added the test case.

lukehsiao (approved):

PR #421, opened 2020-05-20
Enabled "Type hints (PEP 484) support for the Sphinx autodoc extension"

HiromuHota:
With this PR, the Sphinx docs would look nicer because type hints migrate from the method signature to each parameter.

Now:
image

With this PR:
image

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,6 +1,10 @@
 Unreleased_
 -----------
 
+Changed
+^^^^^^^
+* `@HiromuHota`_: Enabled "Type hints (PEP 484) support for the Sphinx autodoc extension."
+
 0.8.2_ - 2020-04-28
 -------------------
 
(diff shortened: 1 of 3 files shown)


HiromuHota:
Also, from now on, I'd suggest to remove ":type" information from the docstring, which is redundant because the type hint is already there in the method signature.

In the example below, we can safety remove ":type candidate_classes: list" and ":type parallelism: int" from the docstring.
image

HiromuHota:
Enabling "sphinx_autodoc_typehints" generates a lot of deprecation warnings like "RemovedInSphinx40Warning: sphinx.util.inspect.Signature() is deprecated."
This issue has been reported at https://github.com/agronholm/sphinx-autodoc-typehints/issues/133 and is now being addressed by that community.
Thus, I don't think there is much to worry about.

lukehsiao (approved):

PR #424, opened 2020-05-20
mention_extractor.apply with clear=True fails if it's not the first run

HiromuHota:
I think this is a regression caused by #381 .
More specifically, #381 removed `fonduer.utils.utils.get_dict_of_stable_id`, which retrieves contexts that are created previously.

`MentionExtractorUDF.apply` checks if a temporary context would conflict with existing contexts. See below:
https://github.com/HazyResearch/fonduer/blob/0dc064d5d81120141ec9c61dec2abee4ad4b78a4/src/fonduer/candidates/mentions.py#L587-L594

Because `dict_of_stable_id` is initialized with an empty dict (ie `dict_of_stable_id = {}`), it fails to check a temporary context against existing contexts.
Before #381, `dict_of_stable_id` is populated as `dict_of_stable_id = get_dict_of_stable_id(doc)`.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -9,6 +9,8 @@ Fixed
 ^^^^^
 * `@kaikun213`_: Fix bug in table range difference calculations.
   (`#420 <https://github.com/HazyResearch/fonduer/pull/420>`_)
+* `@HiromuHota`_: mention_extractor.apply with clear=True now works even if it's not the first run.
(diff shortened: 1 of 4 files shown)


SenWu:
Can we test this patch with image tutorial?

HiromuHota:
Thanks you for your feedback.
I tested the patch against all the tutorials and passed.
https://github.com/HiromuHota/fonduer-tutorials/runs/708125397

lukehsiao (approved):
LGTM!

PR #427, opened 2020-05-23
Remove ":type" and ":rtype" from the docstring when they are redundant

### LINKED ISSUE #422: Remove redundant ":type" and ":rtype" information from the docstring

See https://github.com/HazyResearch/fonduer/pull/421#issuecomment-631587136.

Remove redundant ":type" and ":rtype" information from the docstring

N/A.

**Additional context**

Now Fonduer has type hints in the method definition, which makes ":type" and ":rtype" in docstring redundant.

HiromuHota:
Close #422.


src/list.txt
```diff
@@ -0,0 +1,58 @@
+fonduer.candidates.candidates
```
lukehsiao, line 1: What's the purpose of this file?

HiromuHota: Oh, that was sort of TODO list. My bad.
Also I'll resolve the conflict.

PR #429, opened 2020-05-27
Add get_bbox() to Sentence and (Temporary)SpanMention, and deprecate bbox_from_span and bbox_from_sentence

HiromuHota:
This PR has two benefits:
1. Type hints become short, hence easier to understand: `Tuple[int, int, int, int, int]` to `Bbox`.
2. No need to import `bbox_from_sentence` and `bbox_from_span`. Just use `sentence.get_bbox()` and `span.get_bbox()`. This is actually the benefit from using OOP.

P.S. There are many more spots where OOP (object-oriented programming) is more suited.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,9 +1,19 @@
 Unreleased_
 -----------
 
+Added
+^^^^^
+* `@HiromuHota`_: Add get_bbox() to :class:`Sentence` and :class:`SpanMention`.
+  (`#429 <https://github.com/HazyResearch/fonduer/pull/429>`_)
+
 Changed
 ^^^^^^^
 * `@HiromuHota`_: Enabled "Type hints (PEP 484) support for the Sphinx autodoc extension."
+
+Deprecated
+^^^^^^^^^^
+* `@HiromuHota`_: Deprecated :func:`bbox_from_span` and :func:`bbox_from_sentence`.
+  (`#429 <https://github.com/HazyResearch/fonduer/pull/429>`_)
 
 Fixed
 ^^^^^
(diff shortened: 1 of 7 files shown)


HiromuHota:
I think I fixed the issue.

SenWu:
Thanks for this PR!

What do you think about having a `fonduer.typing` module to store all Fonduer specific typings? `Bbox` is a good example here.

HiromuHota:
Having a `fonduer.typing` module is a good idea.
IMO, this will have type aliases to make codes more readable.
For example, the following type hints are very lengthy and hard to read.

https://github.com/HazyResearch/fonduer/blob/bd79688230ba11f5ed97e58cc78548de14b2c3db/src/fonduer/parser/visual_linker.py#L31-L33

By defining type aliases like below:

```
Alias1 = List[Tuple[Tuple[int, int], str]]
Alias2 = OrderedDict[Tuple[str, int], Tuple[int, int]]
```

This could become 

```
self.pdf_word_list: Optional[Alias1] = None
self.html_word_list: Optional[Alias1] = None 
self.links: Optional[Alias2] = None 
```

However, `Bbox` is not an alias, hence IMO it is not suited to be placed `fonduer.typing`.
Moreover, I'll be adding methods to `Bbox` like `Bbox.horz_aligned` (superseding `bbox_horz_aligned`) and `Bbox.vert_aligned` (superseding `bbox_vert_aligned`), which makes `Bbox` less suitable in `fonduer.typing`.

HiromuHota:
A good example would be: alias for Throtter

Currently,
https://github.com/HazyResearch/fonduer/blob/bd79688230ba11f5ed97e58cc78548de14b2c3db/src/fonduer/candidates/candidates.py#L65

with
```
Throttler=Callable[[Tuple[Mention, ...]], bool]
```

to 

```
throttlers: Optional[List[Throttler]] = None,
```

PR #230, opened 2019-03-23
 update spacy verison to v2.1.x

SenWu:
Closes #227, #229.

PR #431, opened 2020-06-02
Loose the over-restricted version requirement for spaCy and make it v2.1.X

HiromuHota:
I don't see why spaCy of v2.1.3 or later is required by 5396b5837b11e533193dc3260957f949f920ba2b.
Before adapting unit tests to spaCy v2.2, I'd like to test with spaCy v2.1.0 and include it to make the version requirement more meaningful.


lukehsiao:
This seems like a good plug to make more meaningful commit messages. I think we all make bad ones at times, but hopefully all commits that make it into master have good descriptions that can clarify why changes were made.

HiromuHota:
@lukehsiao thank you for your feedback. I revised the commit. Let me know if I need to revise it again.

lukehsiao:
I found the original PR that changed these requirements. Looks like they used to be some issues (e.g., passing with 2.1.2, but failing with 2.1.3). Did we try these tests with versions earlier than 2.1.3?

https://github.com/HazyResearch/fonduer/pull/230

HiromuHota:
#230 added `doc.is_parsed = True`, which may have fixed some problem.
Thank you for your time, but I'd say we should stop here as it would not be worth spending more time in this subtle thing.
I wanted to clean the version requirement, but I don't have any specific use case to use spacy==2.1.0, 2.1.1, or 2.1.2 right now.
We may revisit this PR in the future, but close it for now.

PR #432, opened 2020-06-02
Support spaCy v2.2

### LINKED ISSUE #384: Upgrade spaCy to v2.2.4 or later for easier installation of Japanese backend

Until v2.2.3, spaCy has used python-mecab3 for Japanese backend, which is cumbersome to install.
As of v2.2.4, spaCy replaced python-mecab3 with fugashi for Japanese, which is easier to install.
https://github.com/explosion/spaCy/releases/tag/v2.2.4

Change Fonduer's dependency on spaCy like `spacy>=2.2.4, <2.3.0`.

N/A

**Additional context**

This is not urgent.
As this shift happened very recently, some bugs may be reported in the near future.
We can wait until this new dependency gets stabilized.

HiromuHota:
This will close #384.

This PR
- Expand the version requirement for spaCy to include v2.2.
- Adapt the unit tests to the language model change from v2.1 to v2.2
- Remove some checks from e2e that are checked elsewhere (ie test_candidates.py and test_parser.py)
- Switch from mecab-python3 to fugashi as an extra dependency for Japanese as spaCy did as of v2.2.4.


tests/e2e/test_e2e.py
```diff
@@ -248,28 +178,21 @@ def test_e2e():
...
 
     # Featurization
     featurizer = Featurizer(session, [PartTemp, PartVolt])
 
     # Test that FeatureKey is properly reset
     featurizer.apply(split=1, train=True, parallelism=PARALLEL)
     assert session.query(Feature).count() == 214
-    assert session.query(FeatureKey).count() == 1260
+    assert session.query(FeatureKey).count() == 1281
+    num_feature_keys = session.query(FeatureKey).count()
```
lukehsiao, line 190: I don't know if queries are cached, but perhaps assign the `num_feature_keys` first (1 query), and then assert `1281` on that so that there is only one query, not two.

tests/e2e/test_e2e.py
```diff
@@ -367,26 +299,27 @@ def test_e2e():
         train=True,
         parallelism=PARALLEL,
     )
-    assert session.query(Label).count() == 6478
+    assert session.query(Label).count() == len(train_cands[0]) + len(train_cands[1])
     assert session.query(LabelKey).count() == 9
+    num_label_keys = session.query(LabelKey).count()
```
lukehsiao, line 304: Same here?

tests/e2e/test_e2e.py
HiromuHota: I don't know if queries are cached, either, but I resolved this issue in 7e6be6f34ff7efa0ad8a67b5134d0e318d851d80 by just reusing query results.

tests/e2e/test_e2e.py
HiromuHota: Thanks.
I fixed other parts as well.

PR #433, opened 2020-06-05
Try to reproduce #12, but can't

### LINKED ISSUE #12: Word mismatch between HTML and PDF for visual linker
In the test [md document](https://github.com/SenWu/fonduer/blob/8e8e4d71b82ecbe4b1069ec7fd6860650d4751b1/tests/data/html_simple/md.html) we use an ordered HTML list, which renders as numbers in the PDF. This is causing a mismatch of words when doing the visual parse. The list of words in that document are

HTML | PDF
:---|:---
Sample | Sample
Markdown | Markdown
This | This
is | is
some | some
basic | basic
, | ,
sample | sample
markdown | markdown
. | .
Second | Second
Heading | Heading
Unordered | Unordered
lists | lists
, | ,
and | and
: | :
One | 1
Two | .
Three | One
More | 2
Blockquote | .
And | Two
bold | 3
, | .
italics | Three
, | More
and | Blockquote
even | And
italics | bold
and | ,
later | italics
. | ,
Even | and
bold | even
strikethrough | italics
. | and
A | later
link | bold
to | .
somewhere | Even
. | strikethrough
Here | .
is | A
a | link
table | to
Name | somewhere
Lunch | .
order | Here
Spicy | is
Owes | a
Joan | table
saag | Name
paneer | Lunch
medium | order
$ | Spicy
11 | Owes
Sally | Joan
vindaloo | saag
mild | paneer
$ | medium
14 | $11
Erin | Sally
lamb | vindaloo
madras | mild
HOT | $14
$ | Erin
5 | lamb
Or | madras
inline | HOT
code | $5
like | Or
var | inline
foo | code
= | like
'", 'var
bar | foo
'", '=
; | &apos;bar&apos;;
. | .
Or | Or
an | an
image | image
of | of
bears | bears
The | The
end | end
... | ...

Notice that in the PDF words, the `1`, `2`, and `3` appear, whereas they do not in the HTML.

Ideally this will be resolved when we switch to `pdftotree` and the new parser.

HiromuHota:
Ultimately would fix #12.

### DIFF
--- a/tests/parser/test_parser.py
+++ b/tests/parser/test_parser.py
@@ -12,6 +12,7 @@ from fonduer.parser.preprocessors import (
     TextDocPreprocessor,
     TSVDocPreprocessor,
 )
+from fonduer.utils.utils_visual import Bbox
 
 
 def get_parser_udf(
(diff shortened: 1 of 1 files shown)


HiromuHota:
Close this as #12 itself won't fix.

PR #434, opened 2020-06-06
Run DocPreprocessor._parse_file only once

HiromuHota:
When `doc_loader` is an instance of `DocPreprocessor` or of its child class, `_parse_file` is executed during iteration (the for-loop in this case).
The current implementation iterates `doc_loader` twice (once for `self.last_docs` and another for `in_queue.put(doc)`.
This PR is to run `_parse_file` only once for better computational efficiency.

### DIFF
--- a/src/fonduer/utils/udf.py
+++ b/src/fonduer/utils/udf.py
@@ -62,9 +62,6 @@ class UDFRunner(object):
         if clear:
             self.clear(**kwargs)
 
-        # Track the last documents parsed by apply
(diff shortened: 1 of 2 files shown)


HiromuHota:
Good suggestion!
In addition to checking `self.last_docs`, I'd check `_parse_file` is executed exactly once.

HiromuHota:
Added a unit test that checks the last_docs and _parse_file is executed exactly once.

PR #436, opened 2020-06-10
Docstrings

senwu:


src/fonduer/utils/utils_table.py
```diff
@@ -11,11 +12,21 @@
...
     # if absolute=False, return the raw value of minimum magnitude difference
     # TODO: move back to efficient implementation once it sees that
     # min_range_diff(3,3,2,3) = 0 return max(0, max(a_end - b_start, b_end -
     # a_start))
+
+    :param a_start: The start index of the first object.
+    :param a_end: The end index of the first object.
+    :param b_start: The start index of the second object.
+    :param b_end: The end index of the second object.
+    :param absolute: Whether use absolute value, defaults to True
```
lukehsiao, line 27: Add period to this an the same param in `min_row_diff` and the other functions.

src/fonduer/utils/utils_table.py
```diff
@@ -50,6 +75,13 @@ def min_axis_diff(
     axis: Optional[str] = None,
     absolute: bool = True,
 ) -> int:
+    """Get the minimum axis difference of two sentences or cells.
+
+    :param a: The first cell or sentence.
+    :param b: The second cell or sentence.
+    :param axis: The axis to calculate the difference, defaults to None
```
lukehsiao, line 82: Period here, too.

src/fonduer/utils/utils_table.py
```diff
@@ -76,6 +122,14 @@ def is_axis_aligned(
     axis: Optional[str] = None,
     spread: List[int] = [0, 0],
 ) -> bool:
+    """Check two sentences or cells are axis-wise aligned.
+
+    :param a: The first cell or sentence.
+    :param b: The second cell or sentence.
+    :param axis: The axis to calculate the alignment, defaults to None
+    :param spread: Row/column difference range, defaults to [0, 0]
```
lukehsiao, line 130: And here. Just check all the params to make sure periods are there.

lukehsiao (changes requested):
Minor nits, otherwise looks good. Anything here is better than what we currently have.

lukehsiao (approved):
LGTM. Should be squashed before merging.

PR #438, opened 2020-06-10
fix: no image in child image node

### LINKED ISSUE #437: No image in the child image node
**Describe the bug**
When the html doesn't have an image, the `_parse_figure` from the `parser.py` is not considering that in some cases the `imgs` list could be empty.

**To Reproduce**
1. Have a html with a figure object that doesn't have an image (eg. `<figure></figure>`)
2. Run the `Parser` with a `HTMLDocPreprocessor` instance
3. You will end up with the exception in `Parser.py (line 278 in apply)`

**Expected behavior**
It should exit the `_parse_figure` in case no figure was found

 - OS: [e.g. Ubuntu 18.04] Ubuntu 18.04.1
 - PostgreSQL Version: [e.g. 9.6] 10.12
 - Poppler Utils Version: [e.g. 0.62.0] 0.62.0-2
 - Fonduer Version: [e.g. 0.6.0] 0.8.3

wajdikhattel:
Closes #437


wajdikhattel:
> Thanks for this fix! Can we have a unit test for this fix as well? You can put the test here: https://github.com/HazyResearch/fonduer/blob/master/tests/parser/test_parser.py

Done

PR #439, opened 2020-06-10
Execute preprocessing and parsing in parallel

### LINKED ISSUE #435: All the documents are loaded onto memory at once during UDFRunner.apply and this leads to a slowness if they exceeds machine's memory size
**Describe the bug**

This is not a bug, but a memory-inefficient computing.
When executing `UDFRunner.apply` (like `Parser.apply`), all the documents are loaded onto memory at once (see the code below) before processing them in mutiple processes parallelly.
https://github.com/HazyResearch/fonduer/blob/8134e4d22861cdc3104d3101a2ec7b021d88eef2/src/fonduer/utils/udf.py#L115-L116

When `len(doc_loader)` is large, each file is large, and/or machine's RAM is small, this leads to memory paging and critically slows down the execution.

**To Reproduce**
1. Process a large set of documents on a machine with small RAM

**Expected behavior**

Load and process a document one at a time.
(When parallelism=N, load and process N documents at a time.)

 - OS: N/A
 - PostgreSQL Version: N/A
 - Poppler Utils Version: N/A
 - Fonduer Version: 0.8.2

**Additional context**

HiromuHota:

Currently, preprocessor and parser are executed in a complete sequential order.
i.e., preprocess N docs (and load them into a queue), then parse N docs.
This has two drawbacks:
  1. the progress bar shows nothing during preprocessing.
  2. the machine RAM has to be large enough to hold N preprocessed docs at a time.

They become more serious when N is large and/or each HTML file is large.

Fix #435 

A clear and concise description of what you propose.

This PR
- places a cap on the `in_queue` so that only a certain number of documents are loaded to `in_queue`.
- executes preprocessor and parser in parallel (ie the main process does preprocessing and child process(es) do parsing in parallel).

## Test plan

For the 1st issue: I manually check the progress bar starts showing progress right after starting `parse.apply`.

## Checklist

### DIFF
--- a/src/fonduer/utils/udf.py
+++ b/src/fonduer/utils/udf.py
@@ -104,17 +104,14 @@ class UDFRunner(object):
 
         # Create an input queue to feed documents to UDF workers
         manager = Manager()
-        in_queue = manager.Queue()
+        # Set maxsize (#435). The number is heuristically determined.
+        in_queue = manager.Queue(maxsize=parallelism * 2)
         # Use an output queue to track multiprocess progress
         out_queue = manager.Queue()
 
         # Clear the last documents parsed by the last run
         self.last_docs = set()
-        # Fill input queue with documents
-        for doc in doc_loader:
-            in_queue.put(doc)
-            self.last_docs.add(doc.name)
-        total_count = in_queue.qsize()
+        total_count = len(doc_loader)
 
         # Create UDF Processes
         for i in range(parallelism):
@@ -131,12 +128,23 @@ class UDFRunner(object):
         for udf in self.udfs:
             udf.start()
 
+        iter = doc_loader.__iter__()
         count_parsed = 0
         while (
             any([udf.is_alive() for udf in self.udfs]) or not out_queue.empty()
         ) and count_parsed < total_count:
+            # Fill input queue with documents but # of docs in queue is capped (#435).
+            if not in_queue.full():
+                try:
+                    doc = next(iter)
+                    in_queue.put(doc)
+                    self.last_docs.add(doc.name)
+                except StopIteration:
+                    pass
+
+            # Get doc from the out_queue and persist the result into postgres
             try:
-                y = out_queue.get(timeout=1)
+                y = out_queue.get_nowait()
                 self._add(y)
                 # Update progress bar whenever an item has been processed
                 count_parsed += 1
@@ -150,6 +158,8 @@ class UDFRunner(object):
                 raise (e)
 
         # Join the UDF processes
+        for _ in self.udfs:
+            in_queue.put(UDF.TASK_DONE)
         for udf in self.udfs:
             udf.join()
 
@@ -198,19 +208,18 @@ class UDF(Process):
         Session = new_sessionmaker()
         session = Session()
         while True:
-            try:
-                doc = self.in_queue.get_nowait()
-                # Merge the object with the session owned by the current child process.
-                # If transient (ie not saved), save the object to the database.
-                # If not, load it from the database w/o the overhead of reconciliation.
-                if inspect(doc).transient:  # This only happens during parser.apply
-                    doc = session.merge(doc, load=True)
-                else:
-                    doc = session.merge(doc, load=False)
-                y = self.apply(doc, **self.apply_kwargs)
-                self.out_queue.put(y)
-            except Empty:
+            doc = self.in_queue.get()  # block until an item is available
+            if doc == UDF.TASK_DONE:
                 break
+            # Merge the object with the session owned by the current child process.
+            # If transient (ie not saved), save the object to the database.
+            # If not, load it from the database w/o the overhead of reconciliation.
+            if inspect(doc).transient:  # This only happens during parser.apply
+                doc = session.merge(doc, load=True)
+            else:
+                doc = session.merge(doc, load=False)
+            y = self.apply(doc, **self.apply_kwargs)
+            self.out_queue.put(y)
         session.commit()
         session.close()
 
--- a/src/fonduer/utils/udf.py
+++ b/src/fonduer/utils/udf.py
@@ -1,7 +1,8 @@
 """Fonduer UDF."""
 import logging
 from multiprocessing import Manager, Process
-from queue import Empty, Queue
+from queue import Queue
+from threading import Thread
 from typing import Any, Collection, Dict, Iterator, List, Optional, Set, Type
 
 from sqlalchemy import inspect
@@ -111,7 +112,6 @@ class UDFRunner(object):
 
         # Clear the last documents parsed by the last run
         self.last_docs = set()
-        total_count = len(doc_loader)
 
         # Create UDF Processes
         for i in range(parallelism):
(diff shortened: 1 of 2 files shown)


HiromuHota:
Tests passed, but extremely slow.
While the pytest step took 18m to finish on Linux on the master branch (8edfacd4eabaa9fd49ac71bd63488322fa4b5fa2), it took 28m on 26ac90f0ed4490e1e3c7b012d61ea3d7319a7ddf.

SenWu:
Great! How about we set `maxsize` as argument in [config](https://github.com/HazyResearch/fonduer/blob/master/src/fonduer/utils/config.py) and user can override it.

Is there a way to automatically estimate the maxsize?

HiromuHota:
Good point. I think the current magic number (`maxsize=parallelism * 2`) is good enough, or at least there would be little benefit from optimizing it.

This `maxsize` is the buffer size between preprocessing and parsing.
Let's assume preprocessing is **slower** than parsing , the `in_queue` is always close to empty and `Parser` will be waiting. This case, there is no point in optimizing `maxsize`. `maxsize=1` would be good enough.
Let's assume preprocessing is **faster** than parsing, the `in_queue` will be piling up, and up to `maxsize`. If `maxsize` is large, it could eat up the whole RAM (#435). Having a cap on queue by setting a low number for `maxsize` can prevents this, but `Preprocessor` will be waiting.
So what should actually be optimized is the speed of both preprocessing and parsing, in other words, they have to be balanced for the best performance and the best use of memory.

If you remember that only one process does preprocessing while multiple child process(es) do parsing in parallel, you may want to parallelize preprocessing too.
But then you would wonder what's the best `parallelism` for preprocessing and one for parsing. This will be a nightmare.
By the time we start thinking about the whole pipeline optimization, we have to leverage parallel/distributed computing frameworks like Dask, PySpark instead of direct use of `multiprocessing`.

HiromuHota:
Regarding performance, the current commit (b01a9d54a05aa51c3bbb2a9fad7c76a3609a7053) took 21m in the pytest step on Linux. I think this is within an allowable margin of error. The performance on GitHub Actions fluctuates anyway.

HiromuHota:
I can clearly confirm that the first issue has been resolved, but still seeing memory issue.
I thought I was seeing a memory issue on spaCy (https://github.com/explosion/spaCy/issues/3618) and (https://github.com/explosion/spaCy/pull/4486), but this issue has been fixed at v2.1.9.

HiromuHota:
The 2nd issue: "the machine RAM has to be large enough to hold N preprocessed docs at a time" is not the case.
I ran only the preprocessing part as:
```
from multiprocessing import Queue

queue = Queue()
i = 0
for doc in doc_preprocessor:
    print(i)
    i += 1
    queue.put(doc)
```

I can see an increase of the memory usage, but this did not kill the python process.

Running the following (preprocessing + parsing), ie `parser.apply(doc_preprocessor)` kills it.
So I suspected a memory leak somewhere in parsing, and I updated spaCy from 2.1.9 to 2.2.4, SQLAlchemy from 1.3.13 to 1.3.17, but still having an issue.

HiromuHota:
Now I have a better picture of the issue.
The memory issue is actually two-hold:
1. Loading all the preprocessed documents onto the memory at a time.
2. Parser itself is very memory-hungry. (e.g., a machine with 1GB of RAM barely processes ONSMS04099-1.html, 1.3MB on disk, in the hardware tutorial).

This PR addresses only the 1st issue.

HiromuHota:
It is typically difficult to test the parallel execution of multiple things (preprocessing and parsing in this case).
I hard-coded `print` in udf.py for testing purpose and got the following result when executing test_e2e.py:

```
Main process put        112823 into in_queue (in_queue.qsize: 0)
1-th worker process got 112823 from in_queue (in_queue.qsize: 0)
Main process put        2N3906-D into in_queue (in_queue.qsize: 0)
0-th worker process got 2N3906-D from in_queue (in_queue.qsize: 0)
Main process put        2N3906 into in_queue (in_queue.qsize: 1)
Main process put        2N4123-D into in_queue (in_queue.qsize: 2)
Main process put        2N4124 into in_queue (in_queue.qsize: 3)
Main process put        2N6426-D into in_queue (in_queue.qsize: 4)
1-th worker process got 2N3906 from in_queue (in_queue.qsize: 4)
Main process put        2N6427 into in_queue (in_queue.qsize: 4)
0-th worker process got 2N4123-D from in_queue (in_queue.qsize: 4)
Main process put        AUKCS04635-1 into in_queue (in_queue.qsize: 4)
1-th worker process got 2N4124 from in_queue (in_queue.qsize: 4)
Main process put        BC182-D into in_queue (in_queue.qsize: 4)
Main process put        BC182 into in_queue (in_queue.qsize: 4)
0-th worker process got 2N6426-D from in_queue (in_queue.qsize: 4)
1-th worker process got 2N6427 from in_queue (in_queue.qsize: 3)
Main process put        BC337-D into in_queue (in_queue.qsize: 4)
0-th worker process got AUKCS04635-1 from in_queue (in_queue.qsize: 4)
Main process put        BC337 into in_queue (in_queue.qsize: 4)
0-th worker process got BC182-D from in_queue (in_queue.qsize: 3)
1-th worker process got BC182 from in_queue (in_queue.qsize: 2)
0-th worker process got BC337-D from in_queue (in_queue.qsize: 1)
1-th worker process got BC337 from in_queue (in_queue.qsize: 0)
```

`in_queue.qsize()` returns "the approximate size of the queue" (https://docs.python.org/3/library/queue.html#queue.Queue.qsize) but you can roughly see that the number of documents in `in_queue` is capped (by 4 in this case) and that preprocessing and parsing are executed in parallel.

SenWu:
Great! Glad to have 1st issue fixed!

For the 2nd issue, what's the bottleneck? Spacy or other parts of the parsing procedure? My guess is that Spacy is the bottleneck here (Correct me If I am wrong.)

Also, one thing I am aware of is that we create Spacy model for parsing each document which is not friendly for memory.

HiromuHota:
Good question. I'd like to address the 3rd issue (Parser itself is very memory-hungry) in a different issue/PR.
This could be spaCy, lxml, SQLalchemy, or the parsed `Document` itself.

IMO, before addressing the 3rd issue, there should be a guideline of how much memory (and PARALLEL) is recommended for how large the HTML file could be.
For example, 1GB+ RAM is a must and 2GB+ is recommended for ~1MB of HTML files and PARALLEL=1.
Another example could be 4GB+ is recommended for ~1MB of HTML files and PARALLEL=2.
This recommendation could also be affected by if `visual=True`, if `lingual=True`, or etc.

SenWu:
Agreed! Let's address the 3rd issue in another PR. Before giving any recommendation for memory usage, I think we would be better to do some tests.

src/fonduer/utils/udf.py
```diff
@@ -131,25 +128,35 @@ def _apply(
...
-            except Empty:
-                # This happens when any child process is alive and still processing.
-                pass
             except Exception as e:
                 # Raise an error for all the other exceptions.
                 raise (e)
 
         # Join the UDF processes
+        for _ in self.udfs:
+            in_queue.put(UDF.TASK_DONE)
```
lukehsiao, line 159: Just to make sure I understand, we need to put as many `UDF.TASK_DONE` into the `in_queue` here so that all of the UDFs have a signal to commit and close, correct?

HiromuHota: Correct. This PR changes how a child process finishes their job.
Before: they finish their job when `in_queue` becomes empty.
After: they finish their job when they receive `UDF.TASK_DONE` signal from `in_queue`.
Thus, we need to put `UDF.TASK_DONE` as many as child processes.

lukehsiao (approved):
This is a great step forward. Looking forward the reduced parser memory usage, too.

senwu (approved):
LGTM

HiromuHota:
For future reference, disabling the global cache for IDs (ie `collect_ids=False`) in lxml may help solve the remaining issue (https://benbernardblog.com/tracking-down-a-freaky-python-memory-leak-part-2/).

PR #449, opened 2020-06-13
Sort by block top, block left, top, then left

HiromuHota:
By #444, the sort keys were block top, block left, top, then **bottom**.
However, it should be block top, block left, top, then **left** as indicated by the following comment.
> sort pdf_word_list by page, block top then block left, top, then left

This PR will fix this issue.

Actually, #444 should have failed. This PR needs a unit test that would fail on #444.

### DIFF
--- a/src/fonduer/parser/visual_linker.py
+++ b/src/fonduer/parser/visual_linker.py
@@ -17,6 +17,22 @@ from editdistance import eval as editdist  # Alternative library: python-levensh
 from fonduer.parser.models import Sentence
 from fonduer.utils.utils_visual import Bbox
 
+# Define a type alias for readability
+# PdfWordId is an ID for a word that is unique within a PDF file.
+# The two elements represent (page_num, i), where page_num is 1-based page number and
+# i is a 0-based unique ID for a word within that page.
+PdfWordId = Tuple[int, int]
+# PdfWord is a type alias for (PdfWordId, str), where the second element is a string
+# repsentation of the word.
+PdfWord = Tuple[PdfWordId, str]
+# Similarly, HtmlWordId is an ID for word that is unique within a HTML file.
+# The two elements represent (sentence.stable_id, i), where i is an ID within a
+# Sentence.
+HtmlWordId = Tuple[str, int]
+# Similar to PdfWord, HtmlWord is a type alias for (HtmlWordId, str), where the second
+# element is a string repsentation of the word.
+HtmlWord = Tuple[HtmlWordId, str]
+
 
 class VisualLinker(object):
     """Link visual information with sentences."""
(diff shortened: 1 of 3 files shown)


HiromuHota:
2ac486d demonstrates #444 does **not** sort `pdf_word_list` by left.

src/fonduer/parser/visual_linker.py
```diff
@@ -16,6 +16,9 @@
 from fonduer.parser.models import Sentence
 from fonduer.utils.utils_visual import Bbox
 
+# Define a type alias for readability
+WordId = Tuple[int, int]
```
SenWu, line 20: Can you add documentation about these two arguments?

SenWu, line 20: The more type aliases we use, the more I think we should have a place to store all these.

src/fonduer/parser/visual_linker.py
```diff
@@ -171,7 +174,7 @@ def _coordinates_from_HTML(
         pdf_word_list = sorted(
             pdf_word_list,
             key=lambda word_id__: block_coordinates[word_id__[0]]
-            + coordinate_map[word_id__[0]][1:3],
+            + (coordinate_map[word_id__[0]].top, coordinate_map[word_id__[0]].left),
```
SenWu, line 177: Can we sort them by all four coordinates? I am thinking there might be cases that you have one phrase being a subset of another phrase.

src/fonduer/parser/visual_linker.py
```diff
@@ -160,7 +163,7 @@ def _coordinates_from_HTML(
                     xmax = int(float(word.get("xmax")))
                     for content in self.separators.split(word.getText()):
                         if len(content) > 0:  # Ignore empty characters
-                            word_id = (page_num, i)
+                            word_id: WordId = (page_num, i)
```
SenWu, line 166: This is a good explanation of WordId.

src/fonduer/parser/visual_linker.py
HiromuHota: Thanks! Resolved by a2bcb86a94f2a00d7a9bbaad73f96973f1c4fd14.

HiromuHota: For the record, this comment is related to https://github.com/HazyResearch/fonduer/pull/429#issuecomment-635218428.

HiromuHota: In general, I agree with having `fonduer.typing` module where you define type aliases.
Let's create an issue to address this.

src/fonduer/parser/visual_linker.py
HiromuHota: I believe this was resolved by a2bcb86a94f2a00d7a9bbaad73f96973f1c4fd14.

src/fonduer/parser/visual_linker.py
HiromuHota: Can you elaborate more on a case you are concerned about? Hopefully image with coordinates?

SenWu: For example, in the following special example, there might be two words `SMBT3904` and `SMBT3904...` that overlap with each other and in this case we want to use all four coordinates for sorting.
Screen Shot 2020-06-15 at 4 08 37 PM

HiromuHota: Thank you for the example.
Can we assume that the two overlapping words (ie `<word />`) are in the same `<line />` element?
If so, I think the two words are already sorted by pdftotext, meaning Fonduer does not have to sort them further.
Let me check if pdftotext sorts `<word />`. I may have to check the source code of it (eg https://gitlab.freedesktop.org/poppler/poppler/-/blob/master/utils/pdftotext.cc).

HiromuHota: Update: the `-layout` option of pdftotext "maintain original physical layout". With this option, I assume the `<word />` and other elements are sorted. I still have to check.

HiromuHota: Not quite sure, but pdftotext with `-layout` option sorts words by **top then left**.
https://gitlab.freedesktop.org/poppler/poppler/-/blob/master/poppler/TextOutputDev.cc#L2327

SenWu: Sound good! We can fix this later if this is an issue.

SenWu:
Can you resolve the conflict so we can merge?

HiromuHota:
I resolved the conflict. This PR is ready to be merged.

This PR is basically a rework of #444, where the sort keys were "block top, block left, top, then **bottom**" instead of "block top, block left, top, then **left**".
Hence I don't think we need to mention this in the CHANGELOG.

PR #458, opened 2020-06-16
Fix the non-deterministic behavior in VisualLinker

### LINKED ISSUE #412: Non-deterministic behavior in featurization
**Describe the bug**
When working with large (~7k docs) corpus of hardware datasheets, extracting multiple relations, we expect that the features for each candidate would be deterministic between each run. Even more so if we have `parallelism=1` set in the Featurizer. However, we find that there can be small (e.g., < 5) differences between feature tables, resulting in slightly different sparse matrices, and thus, slightly different results.

**To Reproduce**
Running on the HACK transistor dataset will reproduce the error. However, it will take a long time, and we haven't been able to get a very minimal example that reproduces the error yet. Attached are two feature table dumps between two different runs with `parallelism=1`. Note that there is only a single difference on line `65454`.

[feature_table.tar.gz](https://github.com/HazyResearch/fonduer/files/4589655/feature_table.tar.gz)

Note that it isn't always one difference, and the difference is not deterministic. The different attached is just an example.

**Expected behavior**
We would expect that these feature tables are identical between runs.

**Error Logs/Screenshots**
For convenience, here is the differing line in screenshot form
image

**Additional context**
If the issue is in the UDF implementation, this might affect the Labeler in addition to the Featurizer, since they share a lot of the UDF code.

HiromuHota:

See #412.

Fix #412.

A clear and concise description of what you propose.

Based on my findings reported at #412, the non-deterministic behaviour in `Featurization` boils down to a non-deterministic behaviour in `VisualLinker`, which is caused by the fact that the order of `doc.sentences` is not deterministic unless `order_by=True` in the `backref` argument below.
https://github.com/HazyResearch/fonduer/blob/246a92dd05884e3a706ba216e6177456d537de13/src/fonduer/parser/models/sentence.py#L248-L252
One trivial way to fix this issue is to add `order_by=True` to the `backref`, however, I just sort `sentences` (by position) at the beginning of `VisualLinker.link` and I think this is good enough to fix #412.

## Test plan

Ideally, I'd like to test if my change prevents #412 from happening.
However, this is not feasible as the issue is **non-deterministic** and does not happen every test run.
Instead, I'd test if the result of visual linker is affected by the order of sentences.

## Checklist

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -43,6 +43,9 @@ Fixed
 * `@HiromuHota`_: Fix the order of args to Bbox.
   (`#443 <https://github.com/HazyResearch/fonduer/issues/443>`_)
   (`#444 <https://github.com/HazyResearch/fonduer/pull/444>`_)
+* `@HiromuHota`_: Fix the non-deterministic behavior in VisualLinker.
+  (`#412 <https://github.com/HazyResearch/fonduer/issues/412>`_)
+  (`#458 <https://github.com/HazyResearch/fonduer/pull/458>`_)
 
 0.8.2_ - 2020-04-28
 -------------------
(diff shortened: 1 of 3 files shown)


senwu (approved):
LGTM.

PR #459, opened 2020-06-17
fix pdf_path cannot be without a trailing slash

### LINKED ISSUE #442: pdf_path cannot be without a trailing slash
**Describe the bug**

`docs_path` and `pdf_path` are typically followed by a slash "/" like below:

```python
docs_path = "tests/data/html/"
pdf_path = "tests/data/pdf/"

doc_preprocessor = HTMLDocPreprocessor(docs_path, max_docs=max_docs)

corpus_parser = Parser(
    session,
    parallelism=PARALLEL,
    structural=True,
    lingual=True,
    visual=True,
    pdf_path=pdf_path,
)
```

While `docs_path` can be without a trailing slash, `pdf_path` cannot.

**To Reproduce**
1. Do not set a trailing slash like `pdf_path = "tests/data/pdf"`
2. Parse

**Expected behavior**

`pdf_path` without a trailing slash works too.

**Error Logs/Screenshots**
If applicable, add error logs or screenshots to help explain your problem.

```
I/O Error: Couldn't open file 'tests/data/pdf112823.PDF': No such file or directory.
```

 - OS: macOS 10.15.4
 - PostgreSQL Version: N/A
 - Poppler Utils Version: N/A
 - Fonduer Version: 0.8.2

senwu:

Closes #442

Use the same method `_get_linked_pdf_path ` to check the pdf file in `is_linkable` and `link` in `VisualLinker`.

## Test plan
Test various doc_path and pdf_path formats.

## Checklist

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -32,6 +32,9 @@ Deprecated
 
 Fixed
 ^^^^^
+* `@senwu`_: Fix pdf_path cannot be without a trailing slash.
+  (`#442 <https://github.com/HazyResearch/fonduer/issues/442>`_)
+  (`#457 <https://github.com/HazyResearch/fonduer/pull/457>`_)
 * `@kaikun213`_: Fix bug in table range difference calculations.
   (`#420 <https://github.com/HazyResearch/fonduer/pull/420>`_)
 * `@HiromuHota`_: mention_extractor.apply with clear=True now works even if it's not the first run.
--- a/src/fonduer/parser/parser.py
+++ b/src/fonduer/parser/parser.py
@@ -262,8 +262,7 @@ class ParserUDF(UDF):
             if self.visual:
                 # Use the provided pdf_path if present
                 self.pdf_path = pdf_path if pdf_path else self.pdf_path
-                self.vizlink.pdf_path = self.pdf_path
-                if not self.vizlink.is_linkable(document.name):
+                if not self.vizlink.is_linkable(document.name, self.pdf_path):
                     warnings.warn(
                         (
                             f"Visual parse failed. "
@@ -274,7 +273,12 @@ class ParserUDF(UDF):
                     )
                 else:
                     # Add visual attributes
-                    [y for y in self.vizlink.link(document.name, document.sentences)]
+                    [
+                        y
+                        for y in self.vizlink.link(
+                            document.name, document.sentences, self.pdf_path
+                        )
+                    ]
             return document
         except Exception as e:
             warnings.warn(
--- a/src/fonduer/parser/parser.py
+++ b/src/fonduer/parser/parser.py
@@ -262,6 +262,7 @@ class ParserUDF(UDF):
             if self.visual:
                 # Use the provided pdf_path if present
                 self.pdf_path = pdf_path if pdf_path else self.pdf_path
+                self.vizlink.pdf_path = self.pdf_path
                 if not self.vizlink.is_linkable(document.name):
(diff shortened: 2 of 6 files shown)


src/fonduer/parser/parser.py
```diff
@@ -262,6 +262,7 @@ def apply(  # type: ignore
             if self.visual:
                 # Use the provided pdf_path if present
                 self.pdf_path = pdf_path if pdf_path else self.pdf_path
+                self.vizlink.pdf_path = self.pdf_path
```
HiromuHota, line 265: Is this line is needed?
`pdf_path` is already set at https://github.com/HazyResearch/fonduer/blob/8edfacd4eabaa9fd49ac71bd63488322fa4b5fa2/src/fonduer/parser/parser.py#L250

src/fonduer/parser/visual_linker.py
```diff
@@ -135,25 +132,36 @@ def _extract_pdf_words(self) -> None:
...
+        return None
 
-        return False
+    def is_linkable(self, filename: str, pdf_path: str = None) -> bool:
+        """Verify that the file exists and has a PDF extension.
+
+        :param filename: The path to the PDF document.
+        :param pdf_path: The path to the PDF documents, if any, defaults to None.
+        """
+        return False if self._get_linked_pdf_path(filename, pdf_path) is None else True
```
HiromuHota, line 164: I think `_get_linked_pdf_path` would better be merged into `is_linkable`.

HiromuHota (commented):
Overall, LGTM.
I left a few minor comments.

src/fonduer/parser/parser.py
SenWu: This is a good point. I think it's important since in the `apply` function we expose one `pdf_path` argument for users which will override the `vizlink.pdf_path` if it's not None.

src/fonduer/parser/visual_linker.py
SenWu: We use `_get_linked_pdf_path ` in two places. One in `link` func and another place is here. I propose to have a private func because I think it would be better to use the same func to check if the file is linkable and retrieve the real file path.

src/fonduer/parser/parser.py
HiromuHota: Ah, you are right.
Can you pass `pdf_path` as an argument to `self.vizlink.link` instead? It is hacky to directly change a class variable and and not thread-safe in this case.

src/fonduer/parser/visual_linker.py
HiromuHota: I thought `_get_linked_pdf_path` was only used by `is_linkable`. Thanks for the clarification.

src/fonduer/parser/parser.py
SenWu: Agreed! Addressed.

CHANGELOG.rst
```diff
@@ -32,6 +32,9 @@ Deprecated
 
 Fixed
 ^^^^^
+* `@senwu`_: Fix pdf_path cannot be without a trailing slash.
+  (`#442 <https://github.com/HazyResearch/fonduer/issues/442>`_)
+  (`#457 <https://github.com/HazyResearch/fonduer/pull/457>`_)
```
HiromuHota, line 37: Minor, but I just found this line should be "(`#459 <https://github.com/HazyResearch/fonduer/pull/459>`_)".

Issue #380, opened 2020-03-24
Boost the code coverage to >95%


SenWu:
Improve the unit tests and boost the code coverage to >95%

PR #472, opened 2020-06-30
Add tests for `matchers.py`

HiromuHota:

Towards #380.

Towards #380.

This PR adds more tests.

## Test plan

This PR adds more tests.

## Checklist

* [ ] All new and existing tests passed.

### DIFF
--- a/tests/candidates/test_matchers.py
+++ b/tests/candidates/test_matchers.py
@@ -1,4 +1,6 @@
 """Fonduer unit tests for matchers."""
+from unittest.mock import Mock
+
 import pytest
 from nltk.stem.porter import PorterStemmer
 
@@ -232,6 +234,15 @@ def test_cancat(doc_setup):
     matcher = Concat(matcher0, matcher1, right_required=False)
     assert set(tc.get_span() for tc in matcher.apply(space.apply(doc))) == {"This is"}
 
+    # Test with permutations=False
+    matcher = Concat(matcher1, matcher0, permutations=False)
+    assert set(matcher.apply(space.apply(doc))) == set()
+    # Test with permutations=True
+    matcher = Concat(matcher1, matcher0, permutations=True)
+    assert set(tc.get_span() for tc in matcher.apply(space.apply(doc))) == {"This is"}
+
+    # TODO: Add a test for ignore_sep=False
+
 
 def test_dictionary_match(doc_setup):
     """Test DictionaryMatch matcher."""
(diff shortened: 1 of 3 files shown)


tests/candidates/test_matchers.py
```diff
@@ -173,3 +196,231 @@ def test_inverse(doc_setup):
...
+    assert set(tc.get_span() for tc in matcher.apply(space.apply(doc))) == {"This is"}
+
+    # Test with permutations=False
+    matcher = Concat(matcher1, matcher0, permutations=False)
+    assert set(matcher.apply(space.apply(doc))) == set()
+    # Test with permutations=True
+    matcher = Concat(matcher1, matcher0, permutations=True)
+    assert set(tc.get_span() for tc in matcher.apply(space.apply(doc))) == {"This is"}
+
+    # TODO: Add a test for ignore_sep=False
```
SenWu, line 244: Are we going to have this test? Thanks!

tests/candidates/test_matchers.py
```diff
@@ -173,3 +196,231 @@ def test_inverse(doc_setup):
...
+        "O",
+        "O",
+        "O",
+        "GPE",
+        "O",
+        "DATE",
+        "O",
+    ]
+    doc.sentences[1].ner_tags = ["O", "O", "O", "O", "O", "ORG", "O"]
+    # TODO: replace "NUMBER" with "CARDINAL" (#473)
```
SenWu, line 361: Let's make this in another PR.
Maybe check this depends on SpaCy version?

HiromuHota: I'd fix this in another PR (which will fix #473).

> Maybe check this depends on SpaCy version?

What do you mean by this?

tests/candidates/test_matchers.py
HiromuHota: I was not going to as I don't know how to use `ignore_sep=False`.
But let me see the code and try to implement it!

HiromuHota: This option dates back to https://github.com/snorkel-team/snorkel/issues/262 and https://github.com/snorkel-team/snorkel-extraction/commit/4eb2907f10fb8652fa16296e625ad1bc47e31b9e.

HiromuHota: I think I found a bug around options `sep` and `ignore_sep`, which is not easy to fix.
Let's keep it TODO here instead of writing tests that just pass and fool us.

PR #484, opened 2020-07-17
Use dict instead of list for much faster lookup

### LINKED ISSUE #483: Unbearable slowness in `Featurizer.get_feature_matrices`
## Description of the bug

`featurizer.get_feature_matrices(train_cands)` does not return within a bearable amount of time.

## To Reproduce
1. Deploy the fonduer-tutorials using docker
2. Upgrade fonduer to 2a49291003c2826835b0b900425d7663f0ecc202 (introduced by #407)
3. Run the hardware tutorial up to 2.2 Multimodal Featurization

## Expected behavior

`featurizer.get_feature_matrices(train_cands)` returns about 1 min or so for hardware tutorial.

## Error Logs/Screenshots

When I forcefully cancel the operation, the following stack trace is shown.

```
KeyboardInterrupt                         Traceback (most recent call last)
<timed exec> in <module>

~/.venv/lib/python3.7/site-packages/fonduer/features/featurizer.py in get_feature_matrices(self, cand_lists)
    303             features.
    304         """
--> 305         return get_sparse_matrix(self.session, FeatureKey, cand_lists)
    306 
    307 

~/.venv/lib/python3.7/site-packages/fonduer/utils/utils_udf.py in get_sparse_matrix(session, key_table, cand_lists, key)
    157             else:
    158                 annotations.append({"keys": [], "values": []})
--> 159         result.append(_convert_mappings_to_matrix(annotations, key_names))
    160     return result
    161 

~/.venv/lib/python3.7/site-packages/fonduer/utils/utils_udf.py in _convert_mappings_to_matrix(mappings, keys)
    186         if mapping:
    187             for key, value in zip(mapping["keys"], mapping["values"]):
--> 188                 if key in keys:
    189                     indices.append(keys_map[key])
    190                     data.append(value)
```

 - OS: debian:buster
 - PostgreSQL Version: 12.1
 - Poppler Utils Version: N/A
 - Fonduer Version: 2a49291003c2826835b0b900425d7663f0ecc202

## Additional context

I think this is a regression caused by #407.
Also this issue is not noticeable unless the feature vector is big (like 30000 dimensions).

HiromuHota:

See #483.

Fix #483.

Use dict (O(1)) instead of list (O(n)) for much faster lookup

## Test plan

Tests in this repository do not use enough documents to demonstrate #483 .
I used the hardware tutorial, which uses 100 docs, to demonstrate the issue (described in https://github.com/HazyResearch/fonduer/issues/483#issuecomment-660363569) and how this PR fixes it (in https://github.com/HazyResearch/fonduer/issues/483#issuecomment-660384328).

## Checklist

### DIFF
--- a/src/fonduer/utils/utils_udf.py
+++ b/src/fonduer/utils/utils_udf.py
@@ -186,7 +186,7 @@ def _convert_mappings_to_matrix(
     for mapping in mappings:
         if mapping:
             for key, value in zip(mapping["keys"], mapping["values"]):
-                if key in keys:
+                if key in keys_map:
                     indices.append(keys_map[key])
                     data.append(value)
         indptr.append(len(indices))


lukehsiao:
This seems like a significant enough speedup (20x!) that it's probably worth mentioning in the changelog. What do you think?

lukehsiao (approved):

HiromuHota:
> This seems like a significant enough speedup (20x!) that it's probably worth mentioning in the changelog. What do you think?

Thanks for your suggestion! But I don't think it is worth mentioning in CHANGELOG as this fixes an issue caused by #407, which has not been released yet. From the perspective of a user who upgrades from v0.8.2, they would not notice this change.

lukehsiao:
Great point.

HiromuHota:
If you wonder the time complexity in Python, see https://wiki.python.org/moin/TimeComplexity.

PR #488, opened 2020-07-25
Initialize Drawing object every page not to carry over drawings to the following pages

HiromuHota:

### Expected behavior
When displaying candidates like below, the candidates are highlighted by surrounding rectangles.

```python
from hardware_utils import entity_to_candidates
vis = Visualizer(pdf_path)
vis.display_candidates(cands)
```

### Actual behavior
When displaying multiple candidates that span multiple pages, rectangles are drawn not on the right page but also on the following pages.

N/A.

Initialize Drawing object every page not to carry over drawings to the following pages

## Test plan

Run the hardware tutorial and see the result manually.

## Checklist


HiromuHota:
The hardware tutorial can demonstrates the issue.
The blue rectangle for "BC182BG" is drawn both on the first page and on the second page.

image
image

HiromuHota:
This is the result with this patch:

image
image

PR #490, opened 2020-07-30
Persist doc only when no error happens during parsing

### LINKED ISSUE #489: "Document "{document.name}" not added to database, because of parse error" but being added
## Description of the bug

When an error happens during parsing, an error message: "Document "{document.name}" not added to database, because of parse error" appears, but the `Document` object itself is added to database before the parser error happens.

## To Reproduce
1. Use the latest commit (1d6771befb95f4ae94f308899633294a003dcfd6)
2. Let a parser fail
3. Execute `session.query(Document).count()` to see how many `Documents` are added to the database.

## Expected behavior

Either one of these behaviors
1. The `Document` is not added.
2. The error message gets corrected like "Document is added but not parsed".

## Error Logs/Screenshots

"Document "{document.name}" not added to database, because of parse error"

 - Fonduer Version: 0.8.3-dev (1d6771befb95f4ae94f308899633294a003dcfd6)

## Additional context

I think this is a regression caused by b1b8d24e1133e2229c5c02c6bdcf9eab36301a55, where `doc: Document` is saved to the database by `session.merge(doc, load=True)`, so this could happen on v0.8.0, v0.8.1, v0.8.2 too.

HiromuHota:

See #489.

Fix #489.

Persist doc only when no error happens during parsing.

## Test plan

Added a test that demonstrates #489.

## Checklist

### DIFF
--- a/src/fonduer/parser/parser.py
+++ b/src/fonduer/parser/parser.py
@@ -141,6 +141,11 @@ class Parser(UDFRunner):
             progress_bar=progress_bar,
         )
 
+    def _add(self, doc: Union[Document, None]) -> None:
+        # Persist the object if no error happens during parsing.
+        if doc:
+            self.session.add(doc)
+
     def clear(self, pdf_path: Optional[str] = None) -> None:  # type: ignore
         """Clear all of the ``Context`` objects in the database.
 
--- a/src/fonduer/parser/parser.py
+++ b/src/fonduer/parser/parser.py
@@ -154,7 +154,7 @@ class Parser(UDFRunner):
         self.session.query(Context).delete(synchronize_session="fetch")
 
     def get_last_documents(self) -> List[Document]:
-        """Return the most recently parsed list of ``Documents``.
+        """Return the most recently successfully parsed list of ``Documents``.
 
         :return: A list of the most recently parsed ``Documents`` ordered by name.
         """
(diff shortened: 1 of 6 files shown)


HiromuHota:
PyTorch 1.6.0 has been released just two days ago.
Mypy complains that "unused 'type: ignore' comment" at `torch.__version__`.
I couldn't track down which change at PyTorch v1.6.0 affects this, but removing that ignore comment satisfies Mypy.

HiromuHota:
Please squash it when merging.

src/fonduer/parser/parser.py
```diff
@@ -141,6 +141,11 @@ def apply(  # type: ignore
             progress_bar=progress_bar,
         )
 
+    def _add(self, doc: Union[Document, None]) -> None:
```
SenWu, line 144: Do we use this function?

HiromuHota: `Parser._add` is called in the main process after getting a result from `ParserUDF.apply` at
https://github.com/HazyResearch/fonduer/blob/5909011e9bdc52825e480f941a0dff4737002b8e/src/fonduer/utils/udf.py#L143-L150

The `UDF._add` has been meant to add features/labels to the database, more specifically by 

https://github.com/HazyResearch/fonduer/blob/5909011e9bdc52825e480f941a0dff4737002b8e/src/fonduer/features/featurizer.py#L248-L252

and 

https://github.com/HazyResearch/fonduer/blob/5909011e9bdc52825e480f941a0dff4737002b8e/src/fonduer/supervision/labeler.py#L309-L311

.
So only `Featurizer._add` and `Labeler._add` have been implemented, but I realized that `Parser._add` can be implemented to store the "transient" `Document` to database.
We don't need `_add` for `MentionExtractor` and `CandidateExtractor` as extracted mentions and candidates will be persisted without explicitly adding them to the session.

PR #506, opened 2020-09-10
Support v2.3.X of spaCy, which includes pretrained models for Chinese and Japanese

HiromuHota:

spaCy (v2.3.X) includes pretrained models for Chinese and Japanese.
I'd like to use those models in Fonduer.

N/A.

Simply expand the compatible version of spaCy.

## Test plan

Changed existing tests to adopt to spaCy v2.3.X.
As spaCy's models are statistical, some parts of output like part-of-speech tags and named entity change when a model changes, and they actually did.
To become less dependent on unstable output, I changed assertion checks to just check some values other than "" are filled-in or not.

Instead of exact check on each value like below:
```
sentence.ner_tags == ["O", "O", "GPE"]
```

Now just to check `ner_tags` have some value other than `""`:

```
all(sentence.ner_tags)
```

## Checklist


lukehsiao (approved):
LGTM

PR #515, opened 2020-09-24
Add a missing requirement for ReadTheDocs (#512)

### LINKED ISSUE #512: ReadTheDocs error
## Description of the bug

Currently, some parts of the document cannot be seen like below:
image

## To Reproduce
1. Go to https://fonduer.readthedocs.io/en/latest/user/features.html

## Expected behavior

For example, the featurization page should look like this:

image

## Error Logs/Screenshots

See above.

N/A.

HiromuHota:

See #512

Fix #512

Add a missing requirement for Read the Docs

## Test plan

Ask @lukehsiao to see how docs is built on ReadTheDocs?

## Checklist

* [ ] All new and existing tests passed.

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -1,2 +1,3 @@
 sqlalchemy
 sphinx_autodoc_typehints
+deprecation


lukehsiao:
I've just enabled [auto building docs for PRs](https://docs.readthedocs.io/en/stable/guides/autobuild-docs-for-pull-requests.html), so that should help us find these issues earlier.

One issue is that sphinx might return success, even though sections are missing. We probably want to turn sphinx warnings into errors.

lukehsiao:
With RTD actually building it directly, I think we can forgo the "test docs build" CI action, since it is hard to recreate the exact RTD environment, and we can just use the exact one directly.

HiromuHota:
> With RTD actually building it directly, I think we can forgo the "test docs build" CI action, since it is hard to recreate the exact RTD environment, and we can just use the exact one directly.

Yeah, we have to be aware that the "test docs build" CI action does not run in the exact same RTD environment. But at the same time, it is very cumbersome and error-prone to look at the built RTD page as a review process. I think the official RTD build should fail on the warnings that we want it to fail.

lukehsiao:
Agreed. I just set this up in #516, and looks good. Shows the errors and allows a contributor to see the build results. If the RTD build succeeds, it also will allow the contributor to view the built docs for the PR. Let me know what you think.

lukehsiao (approved):
LGTM. I just merge that other PR, so you can rebase on master.

lukehsiao (approved):

PR #516, opened 2020-09-24
docs: configure RTD using config file

lukehsiao:
This gives us more visibility into the RTD settings, rather than the
settings being only visible to the account owning the docs (lukehsiao).

In addition, this treats warnings and errors so we can catch PR
documentation issues early. See #515.


lukehsiao:
Note how the RTD check failed (as expected, due to the warnings), and allows anyone to go view the build errors.

HiromuHota (approved):
LGTM.
Based on this, #515 now builds the RTD successfully.

PR #517, opened 2020-09-24
docs: fix epub warning by adding version to conf.py

lukehsiao:
We were previously seeing

```
Running Sphinx v3.2.1
loading translations [en]... done
making output directory... done
WARNING: conf value "version" should not be empty for EPUB3
```

See https://readthedocs.org/projects/fonduer/builds/11957268/


HiromuHota (approved):
LGTM.

PR #519, opened 2020-09-30
Add HOCRDocProprocessor and HocrVisualParser

HiromuHota:

This is the second patch that follows #518 .

N/A.

Add `HOCRDocProprocessor` and `HocrVisualParser`

## Test plan

I added a few real hOCR example files.

## Checklist

### DIFF
--- a/docs/user/parser.rst
+++ b/docs/user/parser.rst
@@ -4,20 +4,24 @@ The first stage of Fonduer_'s pipeline is to parse an input corpus of documents
 into the Fonduer_ data model.
 
 Fonduer supports different file formats: CSV/TSV, TXT, HTML, and hOCR.
-The diagram below illustrates how files in each format are preprocessed and consumed by :class:`Parser`.
-Nodes in dark blue represent original source files.
+The diagram below illustrates how files in each format are preprocessed and consumed by
+:class:`Parser`. Nodes in dark blue represent original source files.
 You have to convert some of them into the formats that Fonduer can consume:
 a scanned document (incl. non-searchable PDF) is OCRed and exported in hOCR,
 a (born-digital) PDF is converted into hOCR using tools like pdftotree_.
+It is also possible to convert PDF into HTML using third-party tools,
+but not recommended (see `Visual Parsers`_).
 
 .. mermaid::
 
     graph LR;
-        Scan[(Scan)]--OCR-->hOCR[(hOCR)];
         CSV[(CSV)]-->CSVDoc;
         TXT[(TXT)]-->TXTDoc;
         HTML[(HTML)]-->HTMLDoc;
-        PDF[(PDF)]--Convert-->hOCR2[(hOCR)];
+        PDF[(PDF)]--Convert-->HTML2[(HTML)];
+        HTML2-->HTMLDoc;
+        PDF--Convert-->hOCR2[(hOCR)];
+        Scan[(Scan)]--OCR-->hOCR[(hOCR)];
         hOCR-->HOCRDoc;
         hOCR2-->HOCRDoc;
     subgraph Fonduer
@@ -72,8 +76,8 @@ e.g., bounding boxes of each word.
 Fonduer can parse visual information only for hOCR and HTML files
 with help of :class:`HocrVisualParser` and :class:`PdfVisualParser`, respectively.
 It is recommended to provide documents in hOCR instead of HTML,
-because :class:`PdfVisualParser` is not always accurate by its nature and assigns
-a wrong bounding box to a word.
+because :class:`PdfVisualParser` is not always accurate by its nature and could
+assign a wrong bounding box to a word.
 (see `#12 <https://github.com/HazyResearch/fonduer/issues/12>`_).
 
 .. mermaid::
--- a/docs/user/parser.rst
+++ b/docs/user/parser.rst
@@ -3,6 +3,35 @@ Parsing
 The first stage of Fonduer_'s pipeline is to parse an input corpus of documents
 into the Fonduer_ data model.
 
+Fonduer supports different file formats: CSV/TSV, TXT, HTML, and hOCR.
+The diagram below illustrates how files in each format are preprocessed and consumed by :class:`Parser`.
+Nodes in dark blue represent original source files.
+You have to convert some of them into the formats that Fonduer can consume:
+a scanned document (incl. non-searchable PDF) is OCRed and exported in hOCR,
+a (born-digital) PDF is converted into hOCR using tools like pdftotree_.
+
+.. mermaid::
+
+    graph LR;
+        Scan[(Scan)]--OCR-->hOCR[(hOCR)];
+        CSV[(CSV)]-->CSVDoc;
+        TXT[(TXT)]-->TXTDoc;
+        HTML[(HTML)]-->HTMLDoc;
+        PDF[(PDF)]--Convert-->hOCR2[(hOCR)];
+        hOCR-->HOCRDoc;
+        hOCR2-->HOCRDoc;
+    subgraph Fonduer
+        CSVDoc(CSVDocPreprocessor)-->parser;
+        TXTDoc(TXTDocPreprocessor)-->parser;
+        HTMLDoc(HTMLDocPreprocessor)-->parser;
+        HOCRDoc(HOCRDocPreprocessor)-->parser;
+        parser(Parser)-->others(..);
+    end
+    classDef source fill:#aaf;
+    classDef preproc fill:#afa;
+    class Scan,CSV,TXT,HTML,PDF source;
+    class HOCRDoc,CSVDoc,TXTDoc,HTMLDoc preproc;
+
 Multimodal Data Model
 ---------------------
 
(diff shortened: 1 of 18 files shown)


HiromuHota:
Unfortunately, this PR would not fix the #12. Moreover, #12 won't be fixed by its nature.
I'd recommend the hOCR format if visual information is to be parsed because bboxes and words are perfectly linked in the hOCR.
If hOCR cannot be used for any reason, use `PdfVisualParser` as a last resort.

SenWu:
Can we get all conflicts resolved first? Thanks!

HiromuHota:
I rebased on the master branch and resolved the conflicts. Thanks.

docs/user/parser.rst
```diff
@@ -3,6 +3,35 @@ Parsing
 The first stage of Fonduer_'s pipeline is to parse an input corpus of documents
 into the Fonduer_ data model.
 
+Fonduer supports different file formats: CSV/TSV, TXT, HTML, and hOCR.
+The diagram below illustrates how files in each format are preprocessed and consumed by :class:`Parser`.
+Nodes in dark blue represent original source files.
+You have to convert some of them into the formats that Fonduer can consume:
+a scanned document (incl. non-searchable PDF) is OCRed and exported in hOCR,
+a (born-digital) PDF is converted into hOCR using tools like pdftotree_.
```
lukehsiao, line 11: I'd suggest "native-digital" rather than "born-digital". But fine either way.

docs/user/parser.rst
```diff
@@ -3,6 +3,35 @@ Parsing
...
+
+.. mermaid::
+
+    graph LR;
+        Scan[(Scan)]--OCR-->hOCR[(hOCR)];
+        CSV[(CSV)]-->CSVDoc;
+        TXT[(TXT)]-->TXTDoc;
+        HTML[(HTML)]-->HTMLDoc;
+        PDF[(PDF)]--Convert-->hOCR2[(hOCR)];
```
lukehsiao, line 20: This is kind of noted below, but should we show here that you can work with visual information using PDF+hOCR, or PDF+HTML? It seems right now to only show PDF+hOCR, unless I'm mistaken.

docs/user/parser.rst
```diff
@@ -40,6 +69,33 @@ Visual Parsers
 
 The following docs describe various visual parsers. They parse visual information,
 e.g., bounding boxes of each word.
+Fonduer can parse visual information only for hOCR and HTML files
+with help of :class:`HocrVisualParser` and :class:`PdfVisualParser`, respectively.
+It is recommended to provide documents in hOCR instead of HTML,
+because :class:`PdfVisualParser` is not always accurate by its nature and assigns
```
lukehsiao, line 75: Out of curiosity, how much better is hOCR than PDF?

docs/user/parser.rst
lukehsiao: Maybe just adding another line that shown PDF -- Convert --> HTML?

tests/data/hocr_simple/md.hocr
```diff
@@ -0,0 +1,297 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
lukehsiao, line 1: This could just be a comment in this PR, but it might be nice to note in the PR how these test files were generated just for future reference.

Like the previously existing ones were written in markdown then used pandoc to generate PDF and HTML.

tests/parser/test_preprocessor.py
```diff
@@ -0,0 +1,19 @@
...
+
+from fonduer.parser.preprocessors.hocr_doc_preprocessor import HOCRDocPreprocessor
+
+
+def test_hocrpreprocessor():
+    """Test hOCRDocPreprocessor."""
+    path = "tests/data/hocr_simple/md.hocr"
+    preprocessor = HOCRDocPreprocessor(path=path)
+    doc = next(iter(preprocessor))
+    assert doc.name == "md"
```
lukehsiao, line 11: Might be nice to have some more specific spot checking of some words or something more detailed besides just the title?

lukehsiao (approved):
Just a few nits.

docs/user/parser.rst
HiromuHota: hOCR should be 100% accurate by its design (otherwise there is an error in implementation) because bboxes and words are completely tied in hOCR. Of course it is not always one-to-one mapping between spacy tokens and hOCR tokens. E.g., `"Apple."` in hOCR is tokenized (by spaCy) into `["Apple", "."]`. So you would have to expect this one-to-multi (and also multi-to-one) to be happening because spaCy and OCR engines tokenize slightly differently.

On the other hand, when you use HTML/PDF and `PdfVisualParser`, it is not the case.
The primary reason for this is that Fonduer can't control how PDF is converted to HTML (and vice verse) and doesn't even know how it is done.
As a result, Fonduer doesn't know the order of words in HTML and can only guess, and ended up fuzzy-matching words between HTML and PDF.

docs/user/parser.rst
HiromuHota: You're correct. PDF+HTML is still an option.
This option is mentioned by fd88351698c3d46df7f0d0dec29b6e4fd691e3c3.

docs/user/parser.rst
HiromuHota: Thanks for your suggestion.
I'd like to keep it just because "born-digital" has its own wikipedia page (https://en.wikipedia.org/wiki/Born-digital).

tests/parser/test_preprocessor.py
HiromuHota: Resolved by d0c96fb02d2708228129535da3db5ccc90f9ef01.

tests/data/hocr_simple/md.hocr
HiromuHota: Totally agreed!

Regarding md.hocr, 
1. Rasterize the `tests/data/pdf_simple/md.pdf` in gray
2. Apply OCR (Tesseract 4.1.1)
More specifically, here is what I did:
```
$ pdftoppm -gray -r 300 md.pdf md
$ tesseract --dpi 300 --psm 12 md-1.pgm md hocr
```

Regarding japan.hocr,
1. Simplied `tests/data/pure_html/japan.html` and saved it in HTML (pasted below).
2. Print the simplified HTML using Google Chrome in PDF.
3. Follow the same steps as md.hocr

```html
<!DOCTYPE html>
<html class="client-nojs" lang="ja" dir="ltr">

<head>
    <title>ジャパン - Wikipedia</title>
</head>

<body class="mediawiki ltr sitedir-ltr mw-hide-empty-elt ns-0 ns-subject page-ジャパン rootpage-ジャパン skin-vector action-view">
<p>「にっぽん」あるいはその異読である「じっぽん」の転訛。当時マルコ・ポーロが辿り着いたと言われる元の言語で「日本国」のことをそのような発音で呼んでいた。</p>

</body>
</html>
```

HiromuHota: Regarding 121.hocr	and 1st.hocr, they were also generated from PDFs.
The dataset is available to download at https://www.consumerfinance.gov/credit-cards/agreements/.
1. Download the Archived Q4-2019 agreements
2. Unzip it and find 
    - "121 Financial Credit Union_Visa Credit Card Agreement.pdf" (for 121.hocr)
    - "1st Advantage Federal Credit Union_Platinum Visa Credit Card Agreement.pdf" (for 1st)
3. For 121.hocr, did the same steps as "md.hocr".

"1st.hocr" is a bit complicated and not really reproducible.
I used pdftotext to extract the text part and used a proprietary tool to recognize tables.
Nevertheless, "1st.hocr" serves just an example hocr that contains tables, so it is easily replaceable with something else.

Furthermore, I edited both hocr to make it minimal but good enough to demonstrate problems that I encountered when I was dealing with real hOCRs generated using Tesseract. These problems include
- Multi-to-one and one-to-multi token mappings
- A single ocrx_word could contain two words
- CJK languages that does not have whitespaces between words

Hope this description serves as a good future reference.

HiromuHota:
@lukehsiao thank you so much for reviewing such a big PR and your comments.
Please review my edits and replies. Feel free to ask any further clarification.

docs/user/parser.rst
lukehsiao: Ha, well I stand corrected. I should start using born-digital!

lukehsiao (approved):
Thanks!

I think @SenWu wanted to take a look to, so I'll wait for him. But LGTM.

docs/user/parser.rst
HiromuHota: It is not wrong. They are are all synonyms.
In fact, ABBYY calls it "true" PDF (https://pdf.abbyy.com/pdf-types/).

SenWu (approved):
LGTM

SenWu:
Let's plan to have a tutorial about this PR. This is a really awesome improvement!!! Thoughts? @HiromuHota @lukehsiao

HiromuHota:
I'll definitely update the existing tutorials.
Rather than adding a new tutorial, I'd like to
- pick one of the existing ones and replace `HTMLDocPreprocessor` and `PdfVisualParser` with `HOCRDocProprocessor` and `HocrVisualParser` if it is just drop-in replacement,
- Or enrich the "Parsing Documents into the Data Model" section of intro/Intro_Data_Model.ipynb.

SenWu:
That works as well. We need to show two things: 1) some basics about what's in it and how to use it, and 2) end to end run with high quality.

HiromuHota:
An update: I've started replacing html with hocr (html -> pdfy -> pdf -> pdftotree -> hocr) in the wikipedia tutorial.
With a minor change at parser, all cells run with no error but #527 up until the LF part, which needs a major change.

PR #526, opened 2020-10-20
Unwrap "ocrx_line" as well as "ocr_line" as Fonduer has no data model 

HiromuHota:
##  Description of the problems or issues

Currently, parser unwraps "ocr_line" as Fonduer has no data model for lines.
However, hOCR could contains "ocrx_line" for lines.
Fonduer should unwrap this element as well for the same reason.

N/A

Unwrap "ocrx_line" as well as "ocr_line" as Fonduer has no data model 

## Test plan

Use pdftotree (https://github.com/HazyResearch/pdftotree/pull/95) to convert md.pdf to md.hocr, which contains "ocrx_line" elements.
And check if md.hocr can be correctly parsed.

## Checklist

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -1,7 +1,7 @@
 black>=18.9b0
 flake8>=3.8.0
 flake8-docstrings
-mypy
+mypy==0.782
 nltk
 isort>=5.0.0
 pre-commit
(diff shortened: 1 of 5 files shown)


HiromuHota:
The mypy complains:
```
src/fonduer/utils/utils_table.py:27: error: Value of type variable "_LT" of "min" cannot be "_T"
src/fonduer/utils/utils_table.py:29: error: Value of type variable "_LT" of "min" cannot be "_T"
```

This https://github.com/python/mypy/issues/9582 could be related.

I will pin mypy at 0.782 for now and address it in a different issue.

lukehsiao (approved):

Issue #494, opened 2020-08-20
parser.apply does not return for a long time even though the progress bar indicates it finishes parsing


HiromuHota:
## Description of the bug

This is not a bug, but a performance issue.
This is not noticeable when parsing a small number of documents, but parser.apply does not return even though the progress bar indicates it finishes parsing a long time ago (1 hour or more ago).

## To Reproduce
1. Parse many documents (my case: ~2500)

## Expected behavior

`parser.apply` returns when the progress bar indicates it finished parsing all the documents.

## Error Logs/Screenshots
If applicable, add error logs or screenshots to help explain your problem.

## Environment (please complete the following information)
 - OS: Debian Buster
 - PostgreSQL Version: 12.1
 - Poppler Utils Version: N/A
 - Fonduer Version: 0.8.3+dev (01e0d9319b523aff7aa7f5c583a9f330b0705ecc)

HiromuHota:
I found it stuck at `self.session.commit()` in `UDFRunner._apply`.

https://github.com/HazyResearch/fonduer/blob/01e0d9319b523aff7aa7f5c583a9f330b0705ecc/src/fonduer/utils/udf.py#L168

HiromuHota:
I think this is a regression from 807a4b8a5960ef8cd11df892f98bee60ecf9a804 (#490).
By this commit, all the changes during `parser.apply` are held in memory until they are flushed out to the database, which could easily overflow the RAM size when many documents are processed.

HiromuHota:
> I think the idea of processing the rows in chunks is to limit the
memory usage of the Python process. Until you call session.flush(),
all your modifications are held in memory. For small numbers of rows
this isn't a problem, but for huge numbers, it might be.

https://groups.google.com/g/sqlalchemy/c/hK4vIziMWOQ/m/98O5d2dqCgAJ

robbieculkin:
Hi, I'm seeing this bug in the max_storage_temp_tutorial.
![Screen Shot 2021-01-25 at 3 29 39 PM](https://user-images.githubusercontent.com/20375994/105778943-3a812100-5f22-11eb-9a8e-3d56579919c6.png)

robbieculkin:
I'm still seeing this issue. It looks like all processes' data are held in memory, then written to the DB using a single process. I'm debugging via `top`, where I see 
1. memory usage is split across parallel processes
2. next, all memory stacks up in a single process
3. lastly, my machine kills the process due to OOM error

[...]

PR #541, opened 2021-05-07
Resolve memory leaks caused by adding and commiting to postgres (related to #494)

YasushiMiyata:

Memory leaks happen when Fonduer adds and commits data to postgres after parsing many documents.
This causes `OOM killer` of Ubuntu and hangup Fonduer process. Please refer to #494.

One of the memory leak problems is because sqlalchemy keeps all data, which are parsed from documents, on memory untill `commit` from `add`.

Add `commit` to immediately after `add` to free memory (L136 of `src/fonduer/parser/parser.py`).
I’m definitely sure that no side effect happens because no processes exist with the data on the memory after `add` and `commit`.
By the way, there are two cases doing `commit` after multiple `add`; one is preparing a rollback of inserting data to postgres, and the other is accelerating insert processes for many small data. In contrast, document data in Fonduer are a few large data.

## Test plan
Run an existing tests (No additional tests).

## Checklist

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -9,6 +9,9 @@ Added
   to support hOCR as input file.
   (`#476 <https://github.com/HazyResearch/fonduer/issues/476>`_)
   (`#519 <https://github.com/HazyResearch/fonduer/pull/519>`_)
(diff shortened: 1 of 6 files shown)


lukehsiao:
Please rebase this off master.

YasushiMiyata:
So this PR failes rebase, I close and create new PR.

